Issue #10, opened 2020-05-07
Make requirements-dev recursively install requirements too


vviers:
IMO the (generally) expected behavior when installing dependencies in a file like `requirements-dev.txt` is that dependencies in `requirements.txt` will also be installed.

As of today, this is not the case.

Issue #9, opened 2020-05-07
Ignore Pipenv files in the repo


vviers:
Pipenv is a fairly common framework for managing venvs in python.
The .gitignore of the repo should contain `Pipfile` and `Pipfile.lock`.

PR #11, opened 2020-05-07
Update project config

vviers:

### Issue

- [x] My PR addresses the following Issues :  #9 and #10

### Description

- add `Pipfile*` to `.gitignore` so that developpers can use Pipenv
- add recursive import of `requirements.txt` inside `requirements-dev.txt` so that all dependencies for developing can be installed in one line

### Tests

- This PR does not need tests.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+-r requirements.txt
 ipdb
 flake8
 nose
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
+-r requirements.txt
```
bibimorlet, line 1: Didn't know that! Great 👍

PR #37, opened 2020-07-20
 Integration Tests: Real world testing with Docker-Compose

Haypierre:
WIP: DO NOT MERGE.

I thought It was a good idea to start writing some integration tests for nck.
Docker-compose seems to be a good tool to address this effort.

Doing a `docker-compose up` will launch a container that embeds a mysql db populated of small testing data and call the tests located at /tests/integration/ in a separate container.

Writing the very first test, I've noticed we can't instantiate a MySQL reader in any way, neither calling its class functions 
because of a NotImplementedError.

In fact, both the init method and the _create_engine class method are relying on the following code.
 ` @staticmethod
    def connector_adaptor():
        raise NotImplementedError`

### Issue

- [ ] My PR addresses the following Issue (If an issue exists in JIRA or Github)

### Description

- [ ] Here are some details about my PR, including example of usage.

### Tests

- [x] My PR add the s the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [ ] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body explains "what" and "why", not "how"

### Documentation

- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain docstrings that explain what it does

### DIFF
--- a/nck/readers/sql_reader.py
+++ b/nck/readers/sql_reader.py
@@ -122,7 +122,7 @@ class SQLReader(Reader):
             )
         else:
             self._query = build_custom_query(
-                self._engine, schema, query, watermark_column, self._watermark_value
+                self._engine, query, watermark_column, self._watermark_value
             )
 
     @staticmethod
--- a/tests/integration/test_mysql_reader.py
+++ b/tests/integration/test_mysql_reader.py
@@ -1,0 +1,34 @@
+import os
+from nck.readers.mysql_reader import MySQLReader
+
+
+class TestMySQlReader:
+
+    # using tests/resources/employees.sql database
+
+    basic_query = "SELECT * FROM firstname"
+
+    # +---------+
+    # | myfield |
+    # +---------+
+    # | Pierre  |
+    # | Julien  |
+    # +---------+
+
+    sql_env = {
+        "user" : os.environ["MYSQL_USER"],
+        "password" : os.environ["MYSQL_PASSWORD"],
+        "host" : os.environ["MYSQL_HOST"],
+        "port": os.environ["MYSQL_PORT"],
+        "database" : os.environ["MYSQL_DATABASE"]
+    }
+
+    sql_env["query"] = basic_query
+
+    mySQLReader = MySQLReader(**sql_env)
+
+    def test_one(self):
+        # MySQLReader.read() returns a generator object we need to extract
+        json_stream = list(self.mySQLReader.read())
+        request_result = list(json_stream[0].readlines())
+        assert request_result == [{'myfield': 'Pierre'}, {'myfield': 'Julien'}]
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,11 @@
+FROM python:3.7
+
+ADD requirements* /nautilus-connectors-kit/
+
+WORKDIR /nautilus-connectors-kit
+
+RUN pip install -r requirements-dev.txt
+
+ADD . /nautilus-connectors-kit
+
+CMD [ "pytest", "tests/integration"]
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,0 +1,32 @@
+version: '3.8'
+
+services:
+    db:
+        image: mysql:latest
+        command: --default-authentication-plugin=mysql_native_password
+        environment: 
+            - MYSQL_USER=test
+            - MYSQL_PASSWORD=secret
+            - MYSQL_ROOT_PASSWORD=root-secret
+            - MYSQL_ALLOW_EMPTY_PASSWORD=yes
+            - MYSQL_DATABASE=employees
+        ports: 
+            - "3306:3306"
+        volumes:
+            - ./tests/resources/employees.sql:/docker-entrypoint-initdb.d/employees.sql
+        container_name: nck_test_mysql_db
+        cap_add: 
+            - SYS_NICE
+    tests:
+        build:
+            context: .
+            dockerfile: Dockerfile.test
+        environment:
+           - MYSQL_DATABASE=employees
+           - MYSQL_HOST=db
+           - MYSQL_PORT=3306
+           - MYSQL_USER=test
+           - MYSQL_PASSWORD=secret
+           - PYTHONPATH=.
+        depends_on:
+            - db
--- a/tests/resources/employees.sql
+++ b/tests/resources/employees.sql
@@ -1,0 +1,7 @@
+USE employees;
+
+GRANT ALL ON employees.* TO 'test';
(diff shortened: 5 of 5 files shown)


tests/integration/test_mysql_reader.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from nck.readers.sql_reader import SQLReader
+
+class TestMySQlReader:
+
+        basic_query = "SELECT * FROM employees"
+
+        # engine = SQLReader._create_engine(
```
Cedric-Magnan, line 8: Remove uninformative python comments if not needed

tests/integration/test_mysql_reader.py
```diff
@@ -0,0 +1,29 @@
...
+        #     "127.0.0.1",
+        #     "3306",
+        #     "test",
+        #     "secret",
+        #     "employees"
+        # )
+
+        engine = SQLReader(
```
Cedric-Magnan, line 16: Move all configurations to environment variables or to an external python config file for less variables dependencies !

Cedric-Magnan: Especially if you already declared the values as environment variables to the mysql container in the docker-compose file ([declare it in a .env file and use the env_file parameter](https://medium.com/better-programming/using-variables-in-docker-compose-265a604c2006#:~:text=If%20we%20declare%20the%20same,declared%20in%20the%20Compose%20file.) in both container declarations instead)

tests/integration/test_mysql_reader.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from nck.readers.sql_reader import SQLReader
+
+class TestMySQlReader:
+
+        basic_query = "SELECT * FROM employees"
```
Cedric-Magnan, line 6: Add a test_basic_query method

Cedric-Magnan (commented):
Try to add a functional test that will use NCK with the MySQL reader because at the moment, the tests don't do much concerning NCK.

Also, make sure the code follows PEP8 principles.
We like to use pylint, flake8 or black :
https://books.agiliq.com/projects/essential-python-tools/en/latest/linters.html

Haypierre:
The test pass if launched locally with MYSQL_HOSTNAME=127.0.0.1
Still struggling to connect the test container to the MySQL container.  It should work with MYSQL_HOSTNAME=db.
Need more investigation.

Haypierre:
`docker-compose up` is now working and will launch the first integration test with MySQL - pretty basic though.

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.8'
+
+services:
+    db:
+        image: mysql:latest
+        command: --default-authentication-plugin=mysql_native_password
+        env_file: .env.test
+        environment:
+            - MYSQL_ROOT_PASSWORD=root-secret
```
Cedric-Magnan, line 9: You can add these variables in the .env.test file

.env.test
```diff
@@ -0,0 +1,6 @@
+MYSQL_DATABASE=employees
+MYSQL_HOST=db
+MYSQL_PORT=3306
```
Cedric-Magnan, line 3: This file should be ignored by the .gitignore. Can you add the dotenv file patterns to the .gitignore file and remove this file from the tracked files ?
No need to squash the commits, these secrets are for test purposes.

nck/readers/sql_reader.py
```diff
@@ -122,7 +122,7 @@ def __init__(
             )
         else:
             self._query = build_custom_query(
-                self._engine, schema, query, watermark_column, self._watermark_value
+                self._engine, query, watermark_column, self._watermark_value
```
Cedric-Magnan, line 125: 👍

aderennes (commented):
Plutôt stylé, on pourrait rajouter d'autres tests plus système à savoir gérer les comportement par exemple

- l'utilisateur spécifié n'existe pas
- le mot de passe spécifié n'est pas bon
- pas d'accès à la table

tests/integration/test_mysql_reader.py
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    sql_env["query"] = basic_query
+
+    mySQLReader = MySQLReader(**sql_env)
+
+    def test_one(self):
+        # MySQLReader.read() returns a generator object we need to extract
+        json_stream = list(self.mySQLReader.read())
+        request_result = list(json_stream[0].readlines())
```
Cedric-Magnan, line 33: Can you test in the read method of the SQLReader to change
yield self._run_query()
into
return self._run_query()

In order to see if the "json_stream[0].readlines()" could be simplified into "json_stream.readlines()"

Cedric-Magnan (commented):
Still need some changes, but it's great that it now works !
Care about your commit messages, give at least some informations about the changes for each commit and do not repeat the same message. If you must repeat the same message, then you probably need to rebase and squash some commits of your commit history.
(nevermind, I just saw you actually squashed them)
https://github.com/artefactory/kode#iv---git-flow

tests/integration/test_mysql_reader.py
aderennes: I disagree, the whole philosophy of NCK is to work with generators, it should be kept that way to keep the software consistent

.env.test
Haypierre: I think we better keep the .env.test in the repo as it facilitates the tests launch.
Otherwise, One should have to know what to provide as HOST_NAME, PORT...
In the future, we may want to test other services, and it will be painful to know what environment they require before being able to launch the integration test suit.
Security is here to protect data we care about, no random test data with container lifetime.

tests/integration/test_mysql_reader.py
Cedric-Magnan: Indeed, but we can still use :
yield from self._run_query()
instead of
yield self._run_query()

.env.test
Cedric-Magnan: You're right about the user being able to perform the tests without having to fill in anything but there is still an issue by having a .env file and a .env.test file that don't include the same variables.
Let's keep it simple and just fill in the variables directly in the docker-compose.yml file then (and we could actually rename it into docker-compose.test.yml)

https://github.com/motdotla/dotenv#should-i-have-multiple-env-files

PR #44, opened 2020-09-01
upgrade DBM v1.1 - update get_query and remove sdf requests

bibimorlet:
### Description

- [x] Upgrade DBM connector to v1.1 (Google DBM API v1 is deprecated on February 2021)

### DIFF
--- a/nck/readers/README.md
+++ b/nck/readers/README.md
@@ -383,7 +383,7 @@ Didn't work? See [troubleshooting](#troubleshooting) section.
 |`--dcm-start-date`|Start date of the period to request (format: YYYY-MM-DD)|
 |`--dcm-end-date`|End date of the period to request (format: YYYY-MM-DD)|
 
-### Google Display & Video 360 Reader
+### Google DoubleClick Manager Reader (DBM)
 
 #### Source API
 
--- a/nck/readers/dbm_reader.py
+++ b/nck/readers/dbm_reader.py
@@ -159,8 +159,8 @@ class DbmReader(Reader):
         logging.info("waiting for query of id : {} to complete running".format(query_id))
         query_infos = self.get_query(query_id)
         if query_infos["metadata"]["running"] or (
-            "googleCloudStoragePathForLatestReport" not in query_infos["metadata"].keys()
-            and "googleDrivePathForLatestReport" not in query_infos["metadata"].keys()
+            "googleCloudStoragePathForLatestReport" not in query_infos["metadata"]
+            and "googleDrivePathForLatestReport" not in query_infos["metadata"]
         ):
             raise Exception("Query still running.")
         else:
@@ -174,7 +174,10 @@ class DbmReader(Reader):
             query_id = query_infos["queryId"]
             query_infos = self._wait_for_query(query_id)
 
-        if query_infos["metadata"].get("googleCloudStoragePathForLatestReport", None):
+        if (
+            "googleCloudStoragePathForLatestReport" in query_infos["metadata"]
+            and len(query_infos["metadata"]["googleCloudStoragePathForLatestReport"]) > 0
+        ):
             url = query_infos["metadata"]["googleCloudStoragePathForLatestReport"]
         else:
             url = query_infos["metadata"]["googleDrivePathForLatestReport"]
--- a/nck/readers/dbm_reader.py
+++ b/nck/readers/dbm_reader.py
@@ -114,7 +114,7 @@ class DbmReader(Reader):
         if query_id:
             return self._client.queries().getquery(queryId=query_id).execute()
         else:
-            raise ClickException(f"Please provide a 'query_id' in order to find your query")
+            raise ClickException("Please provide a 'query_id' in order to find your query")
 
     def get_existing_query(self):
         query_id = self.kwargs.get("query_id", None)
--- a/nck/readers/dbm_reader.py
+++ b/nck/readers/dbm_reader.py
@@ -162,7 +162,7 @@ class DbmReader(Reader):
             "googleCloudStoragePathForLatestReport" not in query_infos["metadata"].keys()
             and "googleDrivePathForLatestReport" not in query_infos["metadata"].keys()
         ):
-            raise ClickException("Query still running.")
+            raise Exception("Query still running.")
         else:
             return query_infos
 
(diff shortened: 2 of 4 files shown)


nck/readers/README.md
```diff
@@ -383,7 +383,7 @@ Didn't work? See [troubleshooting](#troubleshooting) section.
 |`--dcm-start-date`|Start date of the period to request (format: YYYY-MM-DD)|
 |`--dcm-end-date`|End date of the period to request (format: YYYY-MM-DD)|
 
-### Google Display & Video 360 Reader
+### Google DoubleClick Manager Reader (DBM)
```
benoitgoujon, line 386: Bien vu et ça me fait penser qu'il y a un truc qu'on a pas fait c'est de créer la doc du nouveau DV360 Reader.

benoitgoujon (approved):
Merci @bibimorlet ! Problème de communication, @Haypierre allait le faire mais bon vous vous êtes vus à temps ... On teste de notre côté les différents rapports pour être sûr qu'il n'y ait pas de problèmes de rétrocompatibilité avec l'ancienne version sur nos rapports.

nck/readers/README.md
bibimorlet: Yes en effet, il faudra le rajouter prochainement 😃

aderennes (approved):
👌

bibimorlet:
Yes en effet, on en a discuté avec @Haypierre, c'est tout bon!
Pour les rapports, les filtres ont changé et renvoient des colonnes différentes maintenant, donc il faut bien vérifier les schémas en sortie. 
Par exemple, si j'ai bien saisi:

Avant: Avec un --dbm-filter FILTER_ADVERTISER "XXXXXXX", les colonnes renvoyées étaient

`Advertiser,Advertiser ID,Advertiser Status,Advertiser Integration Code`

Maintenant: Avec un --dbm-filter FILTER_ADVERTISER "XXXXXXX", la colonne renvoyée est:

`Advertiser ID`

N.B: Même comportement à peu près avec les dimensions Campaign, Insertion Order, Line Item, Partner, Creative

nck/readers/dbm_reader.py
```diff
@@ -268,8 +229,6 @@ def read(self):
             data = self.list_query_reports()
         elif request_type == "lineitems_objects":
             data = self.get_lineitems_objects()
-        elif request_type == "sdf_objects":
-            data = self.get_sdf_objects()
         else:
             raise Exception("Unknown request type")
```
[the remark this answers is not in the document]
bibimorlet: Bien pris en compte, thanks!

nck/readers/dbm_reader.py
```diff
@@ -158,8 +158,11 @@ def create_and_get_query(self):
     def _wait_for_query(self, query_id):
         logging.info("waiting for query of id : {} to complete running".format(query_id))
         query_infos = self.get_query(query_id)
-        if query_infos["metadata"]["running"]:
-            raise Exception("Query still running.")
+        if query_infos["metadata"]["running"] or (
+            "googleCloudStoragePathForLatestReport" not in query_infos["metadata"].keys()
+            and "googleDrivePathForLatestReport" not in query_infos["metadata"].keys()
+        ):
+            raise ClickException("Query still running.")
```
benoitgoujon, line 165: Do we really want a `ClickException` to be raised here? I think it might be confusing for the developer because it has nothing to do with the args he or she passed. It is all about the API. So what about a custom exception?

nck/readers/dbm_reader.py
```diff
@@ -171,7 +174,7 @@ def get_query_report_url(self, existing_query=True):
             query_id = query_infos["queryId"]
             query_infos = self._wait_for_query(query_id)
 
-        if query_infos["metadata"]["googleCloudStoragePathForLatestReport"]:
+        if query_infos["metadata"].get("googleCloudStoragePathForLatestReport", None):
```
benoitgoujon, line 177: What about:

```python
if "googleCloudStoragePathForLatestReport" in query_infos["metadata"]:
    url = query_infos["metadata"]["googleCloudStoragePathForLatestReport"]
```

instead.

I think this is the behavior you want and this is slightly more readable.

nck/readers/dbm_reader.py
```diff
@@ -230,7 +233,7 @@ def read(self):
         elif request_type == "lineitems_objects":
             data = self.get_lineitems_objects()
         else:
-            raise Exception("Unknown request type")
+            raise ClickException("Unknown request type")
```
benoitgoujon, line 236: I think this is useless to check that here anyway because a correct request type is forced by Click thanks to the `Click.Choice(POSSIBLE_REQUEST_TYPES)`.

nck/readers/dbm_reader.py
benoitgoujon: Hm unless the case where the key is present but the value empty can happen. Then you need something like:

```python
if (
    "googleCloudStoragePathForLatestReport" in query_infos["metadata"] and len(query_infos["metadata"]. 
   ["googleCloudStoragePathForLatestReport"]) > 0
):
    url = query_infos["metadata"]["googleCloudStoragePathForLatestReport"]
```

This is more explicit that the `get(my_field, None)`.

nck/readers/dbm_reader.py
bibimorlet: I'll put an Exception like before, because here it's only used to call the retry function

nck/readers/dbm_reader.py
bibimorlet: It was the behavior with DBM v1. Now this is a bit weird with DBM v1.1 and apparently there is now only one key between "googleCloudStoragePathForLatestReport" and "googleDrivePathForLatestReport".

And now, the variable query_infos["metadata"]["running"] (which purpose is to tell if the query/report is running or not) is not working very well with DBM v1.1.
For example, if your report is ready (i.e query_infos["metadata"]["running"] == False), you sometimes have to wait a little bit more to retrieve the url thanks to the key "googleCloudStoragePathForLatestReport" because sometimes the key is not in the API response directly.

benoitgoujon: Ok thank you for the context. Very clear. But it was more about the way you wrote the check. 

I find `if query_infos["metadata"].get("googleCloudStoragePathForLatestReport", None)` not explicit and hard to understand.

bibimorlet: Yes I understand, fixes are coming 👍

PR #54, opened 2020-09-30
upgrade facebook-business to v8.0.4

bibimorlet:

### Issue

- Facebook Business API v5.0 deprecated: API calls don't work since 2020-09-28


gabrielleberanger (approved):
Bien vu ! Et c'est tant mieux si la montée en version de l'API n'impacte pas la logique de notre connecteur :blush:

bibimorlet:
Yes nothing changes in the behavior of the connector, but now our API calls will work again 👍

PR #63, opened 2020-11-16
Facebook Reader: asynchronous requests

bibimorlet:
### Description

- Ads Insights request are now asynchronous.
- Handles requests of bigger reports --> avoids the "Error 500: Please reduce the amount of data you're asking"
- Handles requests of more than 1 or 2 days of data --> Faster History requests

### Handling asynchronous requests with Facebook_reader

1) Query ads_insights report on Facebook Marketing API
- retries: 3.

2) Wait for 100% completion of the report:
- Wait for the report to be complete
- retries: Exponential retries between 5s and 5min
- Timeout: 40min

3) Wait for job to be complete:
- Wait for "Job Completed"
--> because the API sends during a brief period of time "Job Running" and "100%" together.
- retries: exponential retries between 10s and 1min
- Timeout: 5min

### Detected weird behaviors of Facebook API:
- During 2), Facebook API sometimes fails or queries report indefinitely. (especially when reports are way too voluminous)
- During 3), Sometimes "Job Running" never changes to "Job Completed"

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
 PROJECT_ID=artefact-docker-containers
 DOCKER_IMAGE=nautilus-connectors-kit-dev
-DOCKER_TAG=v1.1
+DOCKER_TAG=v1.3
 DOCKER_REGISTRY=eu.gcr.io
--- a/nck/readers/facebook_reader.py
+++ b/nck/readers/facebook_reader.py
@@ -314,7 +314,7 @@ class FacebookReader(Reader):
         self._wait_for_complete_report(async_job)
         return async_job
 
-    @retry(wait=wait_exponential(multiplier=60, max=300), stop=stop_after_delay(2400))
+    @retry(wait=wait_exponential(multiplier=5, max=300), stop=stop_after_delay(2400))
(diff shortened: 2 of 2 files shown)


benoitgoujon:
Is it frequent to get the errors you describe? Is there a pattern in these errors like the volume asked, the type of report or ... ?

bibimorlet:
I didn't notice any recurrent pattern with these errors, but:
- It might probably be correlated with the volume of requested data
- The API always sends "Job Running" with "100%" completion during several seconds, so a condition on percent completion is not enough to tell that a job is completed.

benoitgoujon:
> I didn't notice any recurrent pattern with these errors, but:
> 
> * It might probably be correlated with the volume of requested data
> * The API always sends "Job Running" with "100%" completion during several seconds, so a condition on percent completion is not enough to tell that a job is completed.

Do you have a workaround in mind? We can discuss this here.

bibimorlet:
For the moment I have no other workaround than the one presented in the pull request but we can discuss about it of course!

benoitgoujon (approved):

PR #79, opened 2021-01-07
Add mytarget connector

tom-grivaud:

### Issue

- Need to create a connector for mytarget

### Description

- based on an agency id this can retrieve all the informations of the linked campaigns

### Tests

- No tests yet, they're coming.

### Commits

- I have create various functions to first retrieve token and based on a agency
- Store the token in S3 and if the daily max limit of tokens is reached then get the token from s3 
- I have create functions to get the campaign ids, campaign names and the banner ids. For all of them we create function going further than the 20 elements per request limit fixed.


nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+    's3',
+    aws_access_key_id=secret_keys.get('ACCESS_KEY'),
+    aws_secret_access_key=secret_keys.get('SECRET_KEY'),
+    region_name='eu-west-3',
+    config=Config(
+        retries={
+            'max_attempts': 10
+        }
+    )
+)
```
benoitgoujon, line 61: Is it temporary and only for testing purposes?

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+            grant_type='agency_client_credentials',
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            agency_client_name=agency_client
+        )
+        return dict(
+            url=url,
+            headers=headers,
+            data=payload
+        )
```
benoitgoujon, line 97: For dictionary creation, it is [generally advised](https://stackoverflow.com/questions/17097985/dict-vs-in-python-which-is-better/17098017) to use Python's native syntax and not dict.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+        self.agency = agency
+        self.access_token = None
+        self.agency_client_token = None
+
+    def create_agency_token_request(self, agency_client):
+        url = 'https://target.my.com/api/v2/oauth2/token.json'
+        headers = {
+            'Content-Type': 'application/x-www-form-urlencoded',
+            'Host': 'target.my.com'
+        }
```
benoitgoujon, line 86: This can be considered as config. I know in other readers, there are similar hardcoded URLs and headers in class methods but we should avoid that as much as possible. At least, here, we could set the url as a class constant if we don't want to create a new config file.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def retrieve_agency_client_response(self):
+        return requests.post(**self.create_agency_token_request(self.agency)).json()
+
+    def is_agency_client_token_new(self, rsp):
+        return "error" not in rsp.keys()
+
+    def __write_agency_client_token_s3(self, content: Dict[str, str]):
+        s3_resource.Object('token-mytarget', f'{self.agency_client_token}_last_token.json').put(Body=json.dumps(content))
+        logging.info('The token has been uploaded to the bucket.')
```
benoitgoujon, line 107: This code is very specific to your use case. What if a developer who is working in a GCP environment wants to use the MyTarget reader? Why do we store the token? Is it hard to get a new one?

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def set_agency_client_token(self):
+        self.agency_client_token = self.retrieve_agency_client_token()
+
+    def create_campaign_id_request(self, offset: int):
+        url = "https://target.my.com/api/v2/banners.json"
+        headers = {
+            'Authorization': 'Bearer ' + self.agency_client_token['access_token'],
+            'Host': 'target.my.com'
+        }
```
benoitgoujon, line 132: Would be ok to create a function to generate headers.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+        params = dict(offset=offset)
+        return dict(
+            url=url,
+            headers=headers,
+            params=params
+        )
+
+    def get_campaign_id_response(self, offset: int):
+        return requests.get(**self.create_campaign_id_request(offset)).json()
```
benoitgoujon, line 141: Chaining methods like that is powerful but it prevents us from getting the response object that contains interesting information such as HTTP codes. They would be very helpful in case of errors for ex and we should be sure we log errors properly.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def get_all_campaign_id_ex(self):
+        first_ids = self.get_campaign_id_response(0)
+        count = first_ids['count']
+        ids = [first_ids['items']]
+        if count > 20:
```
benoitgoujon, line 147: Why 20?

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def read(self):
+        self.set_agency_client_token()
+        rsp_id = self.get_all_campaign_id_ex()
+        rsp_name = self.get_all_campaign_names_ex()
+
+        campaign_ids = [element['campaign_id'] for element in rsp_id]
+        rsp_banner = self.get_all_banners_all_camp(campaign_ids)
+        rsp_daily_stat = self.get_daily_banner_stat_response(rsp_banner.keys())
+        print(rsp_daily_stat)
```
benoitgoujon, line 249: 👀

benoitgoujon (commented):
This project is now open source and can be used by teams that run this code in different environments. I am sure we can find a way to make this reader less specific to AWS.

nck/readers/mytarget_reader.py
tom-grivaud: Already removed

nck/readers/mytarget_reader.py
tom-grivaud: Didn't know. Going to edit this.

nck/readers/mytarget_reader.py
tom-grivaud: Ok so keep the same functions but just get the url from a config file ? If we use config files it may be more interesting add all the parameters that vary from one function to another such as 'params' or 'payload' dicts right ?

nck/readers/mytarget_reader.py
tom-grivaud: True, going to remove this part and proceed differently.

nck/readers/mytarget_reader.py
tom-grivaud: Already done.

nck/readers/mytarget_reader.py
tom-grivaud: Already removed.

nck/readers/mytarget_reader.py
tom-grivaud: There's a limit set by the api. Yesterday I saw that we can edit this default limit by adding a limit param. I will solve this as soon as I have finished my other tasks.

benoitgoujon: Ok but how can we edit this default limit? If this is a process that each user has to follow, we can't make it a hardcoded value.

nck/readers/mytarget_reader.py
benoitgoujon: Yes absolutely. Url + payloads/headers/... Everything that is static. Even if a dict is populated at runtime we can use template strings.

nck/readers/mytarget_reader.py
```diff
@@ -218,35 +208,104 @@ def get_all_banners_all_camp(self, campaign_ids):
...
+
+    def get_banner_name_response(
+        self,
+        banner_ids: Dict[str, str]
+    ) -> List[Dict[str, Any]]:
+        dict_ban_name = {}
+        for ban_id in banner_ids:
+            time.sleep(1)
+            dict_ban_name[ban_id] = requests.get(**self.get_banner_name_request(ban_id)).json().get('name')
+            print(dict_ban_name[ban_id])
```
benoitgoujon, line 249: 👀

nck/readers/mytarget_reader.py
```diff
@@ -16,51 +16,36 @@
...
 @click.option("--mytarget-agency", required=True)
+@click.option("--mytarget-refresh-token", required=True)
+@click.option("--mytarget-start-date", type=click.DateTime())
+@click.option("--mytarget-end-date", type=click.DateTime())
+@click.option(
+    "--mytarget-date-format",
+    default="%Y-%m-%d",
+    help="And optional date format for the output stream. "
+    "Follow the syntax of https://docs.python.org/3.8/library/datetime.html#strftime-strptime-behavior",
+)
```
benoitgoujon, line 43: For now, we use the JSONStream, so there is no option to customize date format. So, is this parameter still relevant?

nck/readers/mytarget_reader.py
tom-grivaud: Going to take care of this right away.

nck/readers/mytarget_reader.py
```diff
@@ -69,78 +54,68 @@ def __init__(
...
         self.access_token = None
-        self.agency_client_token = None
-
-    def create_agency_token_request(self, agency_client):
-        url = 'https://target.my.com/api/v2/oauth2/token.json'
-        headers = {
-            'Content-Type': 'application/x-www-form-urlencoded',
-            'Host': 'target.my.com'
+        self.agency_client_token = {
+            'access_token': None,
```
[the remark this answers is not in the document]
tom-grivaud: Done.

nck/readers/mytarget_reader.py
```diff
@@ -69,78 +54,68 @@ def __init__(
...
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Host': 'target.my.com'
+            }
+        elif header_type == "authorization":
+            return {
+                'Authorization': 'Bearer ' + self.agency_client_token['access_token'],
+                'Host': 'target.my.com'
+            }
+        else:
+            logging.ERROR("No such kind of header available")
```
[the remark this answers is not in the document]
tom-grivaud: Correct, already edited.

nck/readers/mytarget_reader.py
```diff
@@ -61,11 +62,84 @@ def __init__(
...
+            try:
+                datetime(date.year, date.month, date.day)
+                logging.info(f'CORRECT: Date valid {date}')
+                return True
+            except ValueError as e:
+                raise ValueError(f'The date is not valid : {e}')
+
+        def __check_date_not_in_future(end_date: datetime) -> bool:
+            if end_date <= datetime.now():
+                logging.info(f'CORRECT: end date anterior or equal to current date {datetime.now()}')
```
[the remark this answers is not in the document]
tom-grivaud: Alright so since there are already comments when an error is raised these "positive" logs should be removed right ?

nck/readers/mytarget_reader.py
```diff
@@ -61,11 +62,84 @@ def __init__(
...
+        yield JSONStream(
+            "results_", self.split_content_by_date(complete_daily_content)
+        )
+
+    def __check_date_input_validity(self) -> bool:
+        """The goal of this function is to check the validity of the date input parameters before retrieving the data.
+        """
+
+        def __is_none(date: datetime) -> bool:
+            return date is None
```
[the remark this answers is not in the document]
tom-grivaud: I actually did this to make the code easier to read as for now there are many functions in the file. For now I'll leave it like this if you don't mind and I'll see how it goes for the testing part.

nck/readers/mytarget_reader.py
```diff
@@ -62,31 +57,17 @@ def __init__(
...
-        response_name = self.get_all_campaign_names()
-
-        names_dict = self.convert_list_dicts_names_to_dict(response_name)
-        ids_dict = self.convert_list_dicts_ids_to_dict(response_id)
-        campaign_ids = [element['campaign_id'] for element in response_id]
-
-        rsp_banner = self.get_all_banners_all_camp(campaign_ids)
-        rsp_daily_stat = self.get_daily_banner_stat_response(rsp_banner.keys())
-        rsp_banner_names = self.get_banner_name_response(list(rsp_banner))
+        rsp_daily_stat, names_dict, ids_dict, rsp_banner_names = self.__retrieve_all_data()
```
benoitgoujon, line 70: I may be wrong but I don't see any return value for this method.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+
+    def __is_none(self, date: datetime) -> bool:
+        return date is None
+
+    def __check_validity_date(self, date: datetime) -> bool:
+        try:
+            datetime(date.year, date.month, date.day)
+            return True
+        except ValueError as e:
+            raise ValueError(f'The date is not valid : {e}')
```
benoitgoujon, line 99: This check is unnecessary because the format is specified with the `click.DateTime()` type. So the date will be valid at this step.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+        self.mail = mail
+        self.agency = agency
+        self.agency_client_token = {'refresh_token': refresh_token}
+        self.start_date = kwargs.get('start_date')
+        self.end_date = kwargs.get('end_date')
+        self.date_format = kwargs.get('date_format')
+        self.day_range = kwargs.get('day_range')
+
+    def read(self):
+        if self.__check_date_input_validity():
```
benoitgoujon, line 67: It would be better, if possible, to put date check in the constructor as for other readers. We don't want to create an instance of a class with bad attributes.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+            self.__retrieve_and_set_token()
```
benoitgoujon, line 68: The authentication should be in the constructor too to be consistent with other readers.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+        return date is None
+
+    def __check_validity_date(self, date: datetime) -> bool:
+        try:
+            datetime(date.year, date.month, date.day)
+            return True
+        except ValueError as e:
+            raise ValueError(f'The date is not valid : {e}')
+
+    def __check_date_not_in_future(self, end_date: datetime) -> bool:
```
benoitgoujon, line 101: This check is useful indeed 👍

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+        if end_date <= datetime.now():
+            return True
+        else:
+            raise ValueError(f'The end date {end_date} is posterior to current date {datetime.now()}')
```
benoitgoujon, line 105: (optional) A custom exception would have been neat. Is is always a good practice to have errors with the narrowest scope.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+
+    def __check_both_start_end_valid(
```
benoitgoujon, line 107: Same here, this check is unnecessary because there is the `required=True` option.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+            campaign_id (int, optional): potential campaign id of the request. Defaults to 0.
+            banner_ids (list, optional): potential banner ids list of the request. Defaults to [].
+
+        Returns:
+            List[Dict[str, Any]]: list of dicts resulting from the requests we made to the api
+        """
+        first_elements = self.__get_response(name_content, campaign_id=campaign_id, banner_ids=banner_ids)
+        count = first_elements['count']
+        elements = [first_elements['items']]
+        if count > 20:
```
benoitgoujon, line 179: We could have defined a constant at the top of the file or in the config and call it here. Thus, we would know the meaning of this number. The constant could have a little documentation on top with a link to the MyTarget doc if needed.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+    def get_all_banners_all_camp(self, campaign_ids, banner_ids=[]):
+        all_banner = []
+        for campaign_id in campaign_ids:
+            all_banner.extend(self.get_all_banners_one_camp(campaign_id, banner_ids=banner_ids))
+        return dict(all_banner)
+
+    def get_all_banners_one_camp(self, campaign_id, banner_ids=[]):
+        all_banners = self.__get_all_results('get_banner_ids', campaign_id=campaign_id, banner_ids=banner_ids)
+        filtered_banners = [(ban['id'], ban['campaign_id']) for ban in all_banners if ban['moderation_status'] == 'allowed']
+        return filtered_banners
```
benoitgoujon, line 195: (optional) `get_all_banners_one_camp` could return a generator. Thus, we would have been able to do a simple list comprehension in `get_all_banners_all_camp`. It would have been more efficient and easily readable.

nck/readers/mytarget_reader.py
```diff
@@ -0,0 +1,408 @@
...
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Host': 'target.my.com'
+            }
+        elif header_type == "authorization":
+            return {
+                'Authorization': 'Bearer ' + self.agency_client_token['access_token'],
+                'Host': 'target.my.com'
+            }
+        else:
+            logging.error("No such kind of header available")
```
benoitgoujon, line 338: We control the header type, aren't we? So, will this case occur? If no, we can simplify and remove this else block. Otherwise, it would be great to log the `header_type`

nck/utils/secret_retriever.py
```diff
@@ -0,0 +1,50 @@
+import boto3
```
benoitgoujon, line 1: Do we still need this file?

benoitgoujon (commented):
Good to see the first tests. Not easy to make something simple. A lot of comments are optional.

Also, don't forget the documentation 🙂

benoitgoujon (approved):
LGTM 👍

Issue #82, opened 2021-01-14
As a User, I can launch a request on any reader using a pre-defined date_range option


gabrielleberanger:
WHY
Currently, only specific readers accept a `--date_range` parameter (it depends of the source API, that offers it or not).

Such a pre-defined `--date_range` option can be very convenient, in particular when conducting tests. In this case, you just have to specify `--date-range PREVIOUS_WEEK` for instance, instead of providing a specific `--start-date` and `--end-date` for your request.

HOW
Make the `--date_range` option available to all readers, including when it is not initially provided by the source API.

benoitgoujon:
Also see issue #67 for a description of how each reader handles date parameters.

Issue #67, opened 2020-12-16
As a User, I can launch a request knowing that all readers follow a common behavior on date parameters


gabrielleberanger:

ali-artefact:
Hello !

So I've been working a bit on this issue. I have taken a look at all the readers in order to understand how they work and how they manage the date range option. First, here is a basic overview : 

**- Adobe Reader:**
    
- There seems to be a problem with set_date_range_report_desc function:
There is a condition « if » where we carry out a test on the « date_range » param, except there is no date_range param informed ever in the whole @click part or am I missing something ? If not, it seems like the options « start_date » and « end_date » would never be used in any case because never considered.

**- Adobe Reader v2:**

- The possibility to inform a generic choice of date like « PREVIOUS_DAY » is not offered. Only start date and end date can be (and must be) informed

**- DBM Reader :**

- There is a rule regarding the start date and end date that set it to a default value that doesn’t exist in other readers :
```
default_start_date = datetime.date.today() - datetime.timedelta(days=2) 
default_end_date = datetime.date.today()
```
- The day range has a value defaulted to « LAST_7_DAYS ».
- Additionnal parameter add-date-to-report : « Sometimes the date range on which metrics are computed is missing from the report. If this option is set to True, this range will be added."

**- DCM Reader :**

- The possibility to inform a generic choice of date like « PREVIOUS_DAY » is not offered.
- The functions that manage date range is defined in an other module in : nck/clients/dcm_client.py

**- Facebook Reader :**

- The custom dates are directly drawn from the Facebook api.

**- Ga Reader :**

- The custom dates (day range) are defined and the default value is « last_7_days ».
- Otherwise the date system works the same as Facebook Reader.

**- Googleads reader :**

- The custom dates are defined in nck.helpers.googleads_helper
- There is an additional parameter : date_range_type that can take different values (« CUSTOM_DATE » or « YESTERDAY », « LAST_7_DAYS »…)
- Offers the possibility to configure a custom date (default value) without specifying a start and end date

**- radarly_reader :**

- difficult to understand

**- SA360reader, SearchConsoleReader, ttd_reader, twitter_ready :**

- dates are asked but no date range is built

**- yandex_statistics_reader :**

- This reader is the most complete regarding the date range management, it processes correctly all cases in the function : « _add_custom_dates_if_set ».

So here's a proposition to manage date ranges in a uniform way for all readers : 

[...]

PR #83, opened 2021-01-15
Feat/date range option for all readers

benoitgoujon:
### Issue

My PR addresses the Issue #82 

### Description

This PR brings the skeleton for default date ranges. This PR only implements the feature in one reader: adobe v2 but it will be very easy to add this feature to other readers and I can take care of that.

So, now you can use the `--adobe-2-0-date-range YESTERDAY` option for example to test the feature.

### Tests

My PR adds test for the logic of date start and date stop generation.

### Documentation

- [x] add documentation for adobe reader and a docstring for the function that is exposed for date start and date stop generation

### DIFF
--- a/docs/source/readers.rst
+++ b/docs/source/readers.rst
@@ -128,7 +128,7 @@ Options                             Definition
 ``--adobe-2-0-metric``              Metric to include in the report
 ``--adobe-2-0-start-date``          Start date of the period to request (format: YYYY-MM-DD)
 ``--adobe-2-0-end-date``            Start date of the period to request (format: YYYY-MM-DD)
-``--adobe-2-0-date-range``          Date range. By default, not available in Adobe, so choose among NCK default values.
+``--adobe-2-0-date-range``          Date range. By default, not available in Adobe, so choose among NCK default values: YESTERDAY, LAST_7_DAYS, PREVIOUS_WEEK, PREVIOUS_MONTH, LAST_90_DAYS
 ==================================  =================================================================================================================================================================================
 
 ----------------------
--- a/docs/source/readers.rst
+++ b/docs/source/readers.rst
@@ -128,6 +128,7 @@ Options                             Definition
 ``--adobe-2-0-metric``              Metric to include in the report
 ``--adobe-2-0-start-date``          Start date of the period to request (format: YYYY-MM-DD)
 ``--adobe-2-0-end-date``            Start date of the period to request (format: YYYY-MM-DD)
+``--adobe-2-0-date-range``          Date range. By default, not available in Adobe, so choose among NCK default values.
 ==================================  =================================================================================================================================================================================
 
 ----------------------
--- a/nck/readers/adobe_reader_2_0.py
+++ b/nck/readers/adobe_reader_2_0.py
@@ -37,7 +37,7 @@ from nck.helpers.adobe_helper_2_0 import (
 from nck.readers.reader import Reader
 from nck.streams.json_stream import JSONStream
 from nck.utils.args import extract_args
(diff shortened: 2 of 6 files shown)


nck/readers/adobe_reader_2_0.py
```diff
@@ -143,21 +148,30 @@ def __init__(
...
+            self.end_date = end_date + timedelta(days=1)
+        else:
+            self.end_date = end_date
+        self.date_range = date_range
         self.ingestion_tracker = []
         self.node_values = {}
 
     def build_date_range(self):
-        return f"{self.start_date.strftime(DATEFORMAT)}/{self.end_date.strftime(DATEFORMAT)}"
+        if self.start_date is not None and self.end_date is not None and self.date_range is None:
```
tom-grivaud, line 168: So do we raise an error in case of both start/end dates and date range are given by the user? What about checking if end-start = date range in case all parameters are given ?
Btw seems like the same error would occur if only one of the start/end date is None and the other one is not.

tom-grivaud (commented):
Pretty neat. 👍 
The next steps would be to add this in a util or another folder so that we can use this logic for other readers right ? Like just build the logic using start-date, end-date and date-range as optional parameters and deal with all the cases possible. Let me know what you think.

nck/readers/adobe_reader_2_0.py
benoitgoujon: Yes correct. This is documented in the issue #67. We do this to avoid conflicts between a date range and 2 dates. For example, if the user set a date start and a date stop but also a date range to LAST_7_DAYS, what would we do? We would have to give the priority to one parameter on the other but it would be implicit.

Your suggestion would work but it would make things more difficult imo, so I prefer letting things like that.

benoitgoujon:
> Pretty neat. 👍
> The next steps would be to add this in a util or another folder so that we can use this logic for other readers right ? Like just build the logic using start-date, end-date and date-range as optional parameters and deal with all the cases possible. Let me know what you think.

Yes, part of the code is already in a utils file. The only thing we have to duplicate for now is the input validation but this is the subject of the other PR on date management: #67

nck/readers/adobe_reader_2_0.py
tom-grivaud: Yes I agree, this is enough for now. The value/time ratio for the feature I mentioned is too low to consider it for now.

tom-grivaud (approved):

gabrielleberanger (approved):
Thank you Benoit, you implemented this feature very smartly! The logic can be easily replicated on other readers. I also agree with the way you chose to handle exceptions: the rest of the job will be done as part of the second PR about dates.

I was just concerned on how your feature would interact with the little nuance of the Adobe API about end dates (we have to increment them by one day to include them in the requested time period), but I saw that you already handled it. All good then! :ok_hand:

docs/source/readers.rst
```diff
@@ -128,6 +128,7 @@ Options                             Definition
 ``--adobe-2-0-metric``              Metric to include in the report
 ``--adobe-2-0-start-date``          Start date of the period to request (format: YYYY-MM-DD)
 ``--adobe-2-0-end-date``            Start date of the period to request (format: YYYY-MM-DD)
+``--adobe-2-0-date-range``          Date range. By default, not available in Adobe, so choose among NCK default values.
```
gabrielleberanger, line 131: Maybe we should just document somewhere what are these NCK default values (once for all readers) ?

benoitgoujon: Yes you are right. For now, as it is implemented in only 1 reader, I've added them directly in the doc of this first reader. Once the feature extended to other readers, we will refactor the doc.

Issue #81, opened 2021-01-14
Fix needed: the default .csv field_size_limit is exceeded while making requests with the DV360 reader


tom-grivaud:
ERROR AND WHY :
While collecting data from the platform dv360 I encountered this issue : 

```
Traceback (most recent call last):
  File "nck/entrypoint.py", line 86, in <module>
    app()
  File "/.../nautilus-connectors-kit/nautilus-env/lib/python3.8/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/.../nautilus-connectors-kit/nautilus-env/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/.../nautilus-connectors-kit/nautilus-env/lib/python3.8/site-packages/click/core.py", line 1164, in invoke
    return _process_result(rv)
  File "/.../nautilus-connectors-kit/nautilus-env/lib/python3.8/site-packages/click/core.py", line 1101, in _process_result
    value = ctx.invoke(self.result_callback, value,
  File "/.../nautilus-connectors-kit/nautilus-env/lib/python3.8/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "nck/entrypoint.py", line 68, in run
    writer.write(stream)
  File "/.../nautilus-connectors-kit/nck/writers/console_writer.py", line 44, in write
    buffer = file.read(1024)
  File "/.../nautilus-connectors-kit/nck/streams/stream.py", line 114, in readinto
    chunk = self.leftover or encode(next(iterable))
  File "/.../nautilus-connectors-kit/nck/utils/file_reader.py", line 36, in sdf_to_njson_generator
    for line in dict_reader:
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/csv.py", line 111, in __next__
    row = next(self.reader)
_csv.Error: field larger than field limit (131072)
```

From the error message I was able to set a csv.field_size_limit above the 131072 default limit. 

HOW TO FIX IT AND FURTHER INVESTIGATIONS :
By adding this line of code to the file_reader.py the error vanished and I was able to get my result prompted on the console.

The line is the following and was added to the nck/utils/file_reader.py file(replace 1000000 by another limit to discuss) : 
```
csv.field_size_limit(10000000)
```

[...]

PR #84, opened 2021-01-18
fix: raise csv field limit value to 1 000 000

tom-grivaud:
### Issue

- [x] Fix needed: the default .csv field_size_limit is exceeded while making requests with the DV360 reader #81 

### Description

- [x] Since the csv field limit was the issue here I raised the limit from the default (131K) to 1 000 000 limit.

### Tests

- [ ] It works locally but now we need to check if the perf are similar or if there is a significant decrease.

### Commits

- [x] I have set the csv field limit value to 1 000 000 in nck/utils/file_reader.py

### DIFF
--- a/nck/utils/file_reader.py
+++ b/nck/utils/file_reader.py
@@ -22,6 +22,8 @@ import gzip
 import zipfile
 import json
 
+csv.field_size_limit(1000000)
+
 
 def unzip(input_file, output_path):
     with zipfile.ZipFile(input_file, 'r') as zip_ref:


gabrielleberanger (approved):

gabrielleberanger:
Thank you Tom ! 👍

benoitgoujon (approved):
In terms of performance, I don't have fears anymore as it is a max value, so it does not mean the space in memory will increase for all cells. So, LGTM 👍

tom-grivaud:
> In terms of performance, I don't have fears anymore as it is a max value, so it does not mean the space in memory will increase for all cells. So, LGTM 👍

Yes exactly that's, what I thought, it will only reach this value if needed. I will still let you know how it works on my use-case. Maybe it would even be interesting to remove a json if one of the field exceed the limit value instead of increasing the limit.

PR #85, opened 2021-01-22
Improve mytarget connector

tom-grivaud:
### Description

- [x] Allow the collection of budget for campaigns. 
- [x] Adapt the script to retrieve general stats and budgets reports independently
- [x] Simplify the code and remove useless functions
- [x] Adapt the doc to the new parameter

### Tests

- [x] Only did manual test with data checks for now

### Commits

- [ ] First commits refer to the budget data collection
- [ ] Then the simplification of the data collection process
- [ ] Eventually the doc edits to add the new parameter

### DIFF
--- a/nck/helpers/mytarget_helper.py
+++ b/nck/helpers/mytarget_helper.py
@@ -1,5 +1,7 @@
 from string import Template
 
+REQUEST_TYPES = ["general", "budget"]
+
 REQUEST_CONFIG = {
     'refresh_agency_token': {
         'url': 'https://target.my.com/api/v2/oauth2/token.json',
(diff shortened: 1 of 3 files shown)


nck/helpers/mytarget_helper.py
```diff
@@ -1,4 +1,5 @@
-from string import Template
+
+REQUEST_TYPES = ["general", "budget"]
```
benoitgoujon, line 2: I think "performance" would have been a better name than "general". This is closer to the names used by the business.

benoitgoujon (approved):
Very useful to also have the budgets 👍

nck/helpers/mytarget_helper.py
tom-grivaud: Np, will do right away

PR #92, opened 2021-02-02
fix: adjust click args

tom-grivaud:
### Description

- [x] Add a date range arg and convert it to start-date and end-date.
- [x] Add a filter for banner ids not linked to campaigns.

### Commits

- [x] Add args and conversion to dates and filter banner ids.


benoitgoujon (approved):
Glad to see that we can reuse the date utils and implement them easily in other readers 👍

PR #95, opened 2021-02-08
Improve date handling

ali-artefact:


nck/utils/date_handler.py
```diff
@@ -37,6 +37,21 @@ def __get_previous_month_dates(current_date: date) -> Tuple[date, date]:
 }
 
 
+def check_date_range_definition_conformity(start_date: date, end_date: date, date_range: str):
+    if start_date is None and end_date is None and date_range is None:
+        raise Exception(
+            "You must at least define a couple \
+                        start-date/end-date or a date-range"
+        )
+    elif (start_date is None and end_date is None) or (start_date is None or end_date is None):
```
tom-grivaud, line 46: Hmm, not sure about this condition. Basically you want to make sure that you either have both start date and end date or none of them right ? Therefore you need to raise an exception for the cases where you have only one of the two which is None and the other one who is not None. Example of cases which will raise an exception : {start_date : None, end_date: 2020-06-01} or {start_date : 2020-06-01, end_date: None}. But here you will raise an exception for a case where both are None which can be a correct case if you only have a date range. Basically your condition needs to be an xor.

tom-grivaud (commented):
I think the core function of this PR is not ready for now. I think tdd would have been interesting for that one and not too complicated. Any tests coming for this specific function ?

nck/utils/date_handler.py
tom-grivaud: xor on start-date/end-date but I may be wrong. Let me know what you think about it.

benoitgoujon:
> I think the core function of this PR is not ready for now. I think tdd would have been interesting for that one and not too complicated. Any tests coming for this specific function ?

I agree with Tom, we absolutely need a test for this function.

nck/utils/date_handler.py
ali-artefact: Yes you were absolutely right, that was a mistake. I think I have taken into account what you said in the latest version of the branch.

tests/utils/test_date_handler.py
```diff
@@ -47,3 +47,17 @@ def test_get_previous_month_dates_if_random_month_of_the_year(self):
...
+            (None, date(2021, 1, 12), "YESTERDAY"),
+            (date(2021, 1, 12), None, None),
+            (date(2021, 1, 12), None, "YESTERDAY"),
+            (None, None, None),
+            (date(2021, 1, 12), date(2021, 1, 31), "YESTERDAY"),
+        ]
+    )
+    def test_check_date_range_definition_conformity(self, start_date, end_date, date_range):
+        with self.assertRaises(Exception):
+            check_date_range_definition_conformity(start_date, end_date, date_range)
```
benoitgoujon, line 63: You also need to test the case where the dates are correctly defined. Just check that the function returns `None`.

nck/utils/date_handler.py
```diff
@@ -38,13 +38,13 @@ def __get_previous_month_dates(current_date: date) -> Tuple[date, date]:
 
 
 def check_date_range_definition_conformity(start_date: date, end_date: date, date_range: str):
-    if start_date is None and end_date is None and date_range is None:
+    if (start_date is None and end_date is None) and date_range is None:
         raise Exception(
```
benoitgoujon, line 42: It would be better to raise a custom exception (a class that inherits from the class `Exception`) that would be more specific and would give more info to the developer.

benoitgoujon (dismissed):
LGTM 👍

benoitgoujon:
Tests fail, can you have a look @ali-artefact please? Then, it will be ready to merge. You resolved conflicts 👍

ali-artefact:
I want to bring to your attention that I have modified one of the core functions of the google ads reader. I'd appreciate if you could take an assiduous look to the modifications I've carried out. All tests pass but as I'm not a user of this reader, I may have compromised its working inadvertently. Thank you !

bibimorlet:
Do you use the max line length = 127 from black? 
Because there are a lot of reformattings and lintings that shouldn't be here.

ali-artefact:
Yes exactly, is that troublesome ?

bibimorlet:
> Yes exactly, is that troublesome ?

There are a lot of lines that are reformatted because of the line length

tests/readers/test_googleads_reader.py
```diff
@@ -219,28 +195,44 @@ def test_validate_correct_id(self, valid_input):
...
+        with self.assertRaises(InconsistentDateDefinitionException):
+            inconsistent_report_definition = {
+                "date_range_type": "YESTERDAY",
+                "start_date": datetime.date(2019, 1, 1),
+                "end_date": datetime.date(2019, 3, 1),
+            }
+            report_definition = self.get_report_definition(inconsistent_report_definition["date_range_type"])
+            temp_kwargs = self.kwargs.copy()
+            temp_kwargs.update(inconsistent_report_definition)
+            GoogleAdsReader(**temp_kwargs).add_period_to_report_definition(report_definition)
```
benoitgoujon, line 227: could we parameterise this test? I have the feeling that we call the same method, we just test it raises a different exception against a different set of parameters. It would improve the readability.

nck/readers/googleads_reader.py
```diff
@@ -246,19 +249,30 @@ def get_report_definition(self):
...
+            logger.info(
+                "Date format used for request : Custom Date Range with\
+            start_date and end_date provided"
+            )
             report_definition["selector"]["dateRange"] = self.create_date_range(self.start_date, self.end_date)
+        elif self.start_date is not None and self.end_date is not None and self.date_range_type != "CUSTOM_DATE":
+            raise InconsistentDateDefinitionException(
+                "You must define either the couple start_date and end_date \
+            or a date_range \(different from CUSTOM_DATE\), but not both"
+            )
```
benoitgoujon, line 266: There is a function named `check_date_range_definition_conformity` that seems to do the same thing. Why not use it here?

ali-artefact: I cannot use this function and that's because of the way google_ads has been coded initially. The parameter date_range_type is always defined (it has a default parameter in the click sequence, and I decided not to touch it).

tests/readers/test_googleads_reader.py
ali-artefact: Yes we can, didn't know we could parametrize with different exceptions, good to know.

nck/readers/googleads_reader.py
benoitgoujon: Ok. I think we should remove the default parameters from all readers as it goes against our will that is to remove all implicit behaviours. So, feel free to remove them and use your function. It is here for that exact purpose.

benoitgoujon (approved):

Issue #93, opened 2021-02-08
As a User, I can provide all arguments of as click options for the GCS Reader, S3 Reader, GCS Writer and BQ Writer


gabrielleberanger:
WHY
The following readers/writers are using implicit `config.<VAR>` variables:
- `gcs_reader`, via the `object_storage_reader`
   - `config.PROJECT_ID`
- `s3_reader`, via the `object_storage_reader`
   - `config.REGION_NAME`, `config.AWS_ACCESS_KEY_ID`, `config.AWS_SECRET_ACCESS_KEY`
- `gcs_writer`
   - `config.PROJECT_ID`
- `bq_writer`
   - `config.PROJECT_ID`

These variables are not provided by the user as click options, but as environment variables. These environment variables are retrieved from the following snippet that appears in the `nck/config.py` module:
```
for key, var in os.environ.items():
    locals()[key] = var
```
This can be quite confusing for the user, and does not match the common NCK logic.

HOW
Explicitly implement these variables as click options

PR #99, opened 2021-02-10
fix : corrected unworking storageobject_reader

### LINKED ISSUE #91: BUG - Fix GCS Reader (deprecated)
WHY
gcs_reader seems to be deprecated.
Its purpose is currently to read structured files like .csv, split it and read it as structured data.
This reader should read files in GCS as raw data in order to transfer it without any process into a NCK writer.

HOW
Fix behavior of the gcs_reader

louisRDSC:
### Issue

- [x] My PR resolves #91

### Description

-  #removed broken and useless code in `nck/readers/objectstorage_reader.py`, 
- Created njson reader in `nck/utils/file_reader.py`
- Moved reader factory from `nck/readers/objectstorage_reader.py` to `nck/utils/file_reader` .

### DIFF
--- a/nck/readers/gcs_reader.py
+++ b/nck/readers/gcs_reader.py
@@ -28,7 +28,7 @@ import urllib
 @click.command(name="read_gcs")
 @click.option("--gcs-bucket", required=True)
 @click.option("--gcs-prefix", required=True, multiple=True)
-@click.option("--gcs-format", required=True, type=click.Choice(["csv", "gz"]))
+@click.option("--gcs-format", required=True, type=click.Choice(["csv", "gz", "njson"]))
 @click.option("--gcs-dest-key-split", default=-1, type=int)
 @click.option("--gcs-csv-delimiter", default=",")
 @click.option("--gcs-csv-fieldnames", default=None)
@@ -39,14 +39,10 @@ def gcs(**kwargs):
 
 class GCSReader(ObjectStorageReader, GoogleBaseClass):
     def __init__(self, bucket, prefix, format, dest_key_split=-1, **kwargs):
-        super().__init__(
-            bucket, prefix, format, dest_key_split, platform="GCS", **kwargs
-        )
+        super().__init__(bucket, prefix, format, dest_key_split, platform="GCS", **kwargs)
 
     def create_client(self, config):
-        return storage.Client(
-            credentials=self._get_credentials(), project=config.project_id
-        )
+        return storage.Client(credentials=self._get_credentials(), project=config.project_id)
 
     def create_bucket(self, client, bucket):
         return client.bucket(bucket)
--- a/nck/utils/file_reader.py
+++ b/nck/utils/file_reader.py
@@ -15,7 +15,6 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-from enum import Enum
 import csv
 import codecs
 import gzip
@@ -26,7 +25,7 @@ csv.field_size_limit(1000000)
 
 
 def unzip(input_file, output_path):
-    with zipfile.ZipFile(input_file, 'r') as zip_ref:
+    with zipfile.ZipFile(input_file, "r") as zip_ref:
         zip_ref.extractall(output_path)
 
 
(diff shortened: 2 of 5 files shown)


gabrielleberanger:
Good job, thank you Louis ! 👍 

Did you also have the chance to test the S3 reader ? If needed, I believe that you can ask @tom-grivaud for temporary accesses, as he is working on a project deployed on AWS.

nck/readers/gcs_reader.py
```diff
@@ -39,14 +39,10 @@ def gcs(**kwargs):
...
-        super().__init__(
-            bucket, prefix, format, dest_key_split, platform="GCS", **kwargs
-        )
+        super().__init__(bucket, prefix, format, dest_key_split, platform="GCS", **kwargs)
 
     def create_client(self, config):
-        return storage.Client(
-            credentials=self._get_credentials(), project=config.project_id
-        )
+        return storage.Client(credentials=self._get_credentials(), project=config.project_id)
```
bibimorlet, line 45: I definitely don't like this way to call the GCS project_id. Could we find another to call it?

bibimorlet:
Hey
I tried this new version of the reader, good job.

There are two things I'd like to discuss with you:
- If think that the `config.project_id` in the GCSReader could be removed. We could for example call the project_id with a new click_parameter or another way.
- Can you add some tests for your reader? in order to check if all the desired use cases are validated.

Thanks 😃

bibimorlet (changes requested):
Do we add some tests?

nck/utils/file_reader.py
```diff
@@ -53,20 +52,42 @@ def format_csv_fieldnames(csv_fieldnames):
...
-        raise TypeError(
-            f"The CSV fieldnames is of the following type: {type(csv_fieldnames)}."
-        )
+        raise TypeError(f"The CSV fieldnames is of the following type: {type(csv_fieldnames)}.")
     assert isinstance(_csv_fieldnames, list)
     return _csv_fieldnames
 
 
-class CSVReader(object):
+class FormatReader:
```
benoitgoujon, line 60: Do we really need a class here. Maybe we could keep things simple with just the `create_file_reader` function. What do you think?

tests/readers/test_objectstorage_reader.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@mock.patch("nck.readers.objectstorage_reader.ObjectStorageReader.create_client")
+@mock.patch("nck.readers.objectstorage_reader.ObjectStorageReader.create_bucket")
+@mock.patch.object(ObjectStorageReader, "download_object_to_file", write_to_file)
+@mock.patch.object(ObjectStorageReader, "to_object", mock_to_object)
+@mock.patch.object(ObjectStorageReader, "get_timestamp", mock_get_timestamp)
+@mock.patch.object(ObjectStorageReader, "list_objects", mock_list_objects)
+@mock.patch.object(ObjectStorageReader, "get_key", mock_get_key)
+class ObjectStorageReaderTest(TestCase):
```
benoitgoujon, line 68: I'm wondering if `MagicMock` could have been helpful here. Based on [this example](https://docs.python.org/3/library/unittest.mock.html#quick-guide), it looks like it could have made things a little bit easier.

```python
mock_object_reader = ObjectStorageReader()
mock_object_reader. get_timestamp = MagicMock(return_value=123)
``` 

Not sure it can help but I would dig into that to improve readability.

louisRDSC: I considered something like that but if we want to simulate the reading of more than 1 file, all of them would have the same timestamp. don't you think it can be a problem?

nck/utils/file_reader.py
louisRDSC: yes you are totally right.

tests/readers/test_objectstorage_reader.py
```diff
@@ -0,0 +1,111 @@
...
+    def test_wrong_format(self, a, b):
+        with self.assertRaises(NotImplementedError):
+            reader = ObjectStorageReader(
+                bucket="", prefix=["a"], file_format="txt", dest_key_split=-1, csv_delimiter=",", csv_fieldnames=None
+            )
+            reader
```
bibimorlet, line 74: useless line here 👍

tests/readers/test_objectstorage_reader.py
```diff
@@ -0,0 +1,111 @@
...
+        nb_file = len(list(reader.read()))
+        """check if filter njson with prefix ["b"]"""
+        self.assertEqual(nb_file, 1)
+
+    def test_ObjectStorageReader_read_all_file_CSV(self, a, b):
+        reader = ObjectStorageReader(
+            bucket="", prefix=["a"], file_format="csv", dest_key_split=-1, csv_delimiter=",", csv_fieldnames=None
+        )
+
+        expected = [{"a": "4", "b": "5", "c": "6"}, {"a": "7", "b": "8", "c": "9"}]
```
bibimorlet, line 96: So we read a .csv file and we write a .njson output?
Simple discussion here

Maybe with csvWriters or others that will be developed soon it will be more consistent

nck/readers/gcs_reader.py
benoitgoujon: There's already an issue on this topic: #93 .

bibimorlet: Yes, I forgot to resolve this conversation, thanks!

benoitgoujon (approved):
LGTM 👍

bibimorlet (approved):
Everything works fine for me, Thanks!

PR #101, opened 2021-02-11
feat/add customer id granularity for mcc use

tom-grivaud:
### Description

- [x] The number of advertiser (customer) ids which can be linked to an account is limited to 20. To bypass this limitation is it advised to use an MCC id which is call manager id here. From this MCC you can access all the data for the customer ids which are "bellow" this MCC. For example if ids "1" and "2" are linked to the MCC you can use the MCC creds to access both of these ids. For now we can do so by specifying a manager id and we get all the ids linked to the MCC. Nevertheless we encountered a limitation in one of our use cases for our client: what if we only want to retrieve a specific subset of ids under the mcc ?

To solve this issue, I created a method to update the customer ids which will be used to query the API. This update will compare the customer ids valid for the MCC to the ones we provided.

Cases covered:
1) (New case) Manager id and customer ids provided: 
  - A warning will appear in the case where some ids provided are not linked to the mcc.
  - If none of the ids provided are linked to the mcc we raise a dedicated exception.
  - If one or more ids provided are linked to the mcc we set them as the customer ids used for the query.
2) (Already existing) Manager id provided but no customer id:
  - If no customer id is given as an input we get all the linked ids as before and use them for the query.
3) (Already existing) No manager id provided but customer ids:
  - We don't update customer ids.

IMPORTANT: @gabrielleberanger do you think this interferes with your current use of the googleads connector for the projects with your client ?

### Tests

- [x] My PR adds the following unit tests : tests for the functions used to get the invalid ids but also the common ids between ids linked to mcc and customer ids provided. => tests/utils/test_list_comparator
- [ ] NOT DONE YET: test the update function ?

### Commits

- [x] Add tests to cover the function
- [x] Add a custom exception
- [x] Create list comparator to get correct ids and warn if some ids are not linked to the mcc
- [x] Create a function to update the customer ids given to the query

### Documentation

- [x] -> See docstrings and tests.
- [x] -> See example below of a case where only one id is linked to the mcc but the other are not. 
Capture d’écran 2021-02-11 à 17 04 36

### DIFF
--- a/nck/readers/googleads_reader.py
+++ b/nck/readers/googleads_reader.py
@@ -31,6 +31,8 @@ from googleads.errors import AdWordsReportBadRequestError
 from nck.readers.reader import Reader
 from nck.utils.args import extract_args
 from nck.utils.retry import retry
+from nck.utils.exceptions import NoValidCustomerIdProvidedInMCC
+from nck.utils.list_comparator import get_intersection_two_lists, get_difference_two_list_order
 from nck.commands.command import processor
 from nck.streams.normalized_json_stream import NormalizedJSONStream
 from nck.helpers.googleads_helper import (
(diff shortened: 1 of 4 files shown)


bibimorlet:
I am maybe wrong but this use case was not handled with the current version of NCK?
in fact, if you need to query N advertisers under a MCC, you are able to if you have access to the MCC.
Why do you add the new use case "Manager id and customer ids provided" ?

tom-grivaud:
> I am maybe wrong but this use case was not handled with the current version of NCK?
> in fact, if you need to query N advertisers under a MCC, you are able to if you have access to the MCC.
> Why do you add the new use case "Manager id and customer ids provided" ?

Hmm maybe I got something wrong then. But how did it work before if you needed to only get a specific subset of ids and not all ids linked to one specific manager id ? From the test I made I wasn't able to retrieve a customized list of ids to retrieve from one manager id. How did you manage to do this with your use case ?

Say one manager id has ids 1, 2 and 3 linked to id. What if you only want the data from ids 1 and 2 using the manager id ? I may have misunderstood the way this part of the connector works...

tom-grivaud:
> > I am maybe wrong but this use case was not handled with the current version of NCK?
> > in fact, if you need to query N advertisers under a MCC, you are able to if you have access to the MCC.
> > Why do you add the new use case "Manager id and customer ids provided" ?
> 
> Hmm maybe I got something wrong then. But how did it work before if you needed to only get a specific subset of ids and not all ids linked to one specific manager id ? From the test I made I wasn't able to retrieve a customized list of ids to retrieve from one manager id. How did you manage to do this with your use case ?
> 
> Say one manager id has ids 1, 2 and 3 linked to id. What if you only want the data from ids 1 and 2 using the manager id ? I may have misunderstood the way this part of the connector works...

Important info I have forgot to mention in this message but which I mentioned in the first sentence fo my PR: we can only access 20 account without using MCC. For now we have a simple account  to which can be linked a maximum of 20 customer ids. That's why need to use a manager id but still need to get a custom subset of customer ids from a mcc.

tom-grivaud:
The value given by this update is that it gives a safety net in case one id is not linked to the MCC (which is common in our use case) instead of getting errors if only one of the ids provided is not correct. Should I close the PR ?

Issue #103, opened 2021-02-17
As a user, I need non-normalized stream in the readers output because we normalize the output at the entrypoint with the --normalize-keys parameter


bibimorlet:
Current behavior:
JSONStream or NormalizedJSONStream are randomly defined in readers as outputs.

Expected behavior:
As a user, I want to have non-normalized streams (JSONStream for example) in the readers because I want to normalize them with the --normalize-keys parameter at the entrypoint.py level if needed.

PR #104, opened 2021-02-17
add normalize-keys at entrypoint

bibimorlet:
### Issue

- [x] My PR introduces the issue #103.
In fact, some readers return normalized streams. They should return non-normalized streams.

### Description

Retrieve --normalize-keys at entrypoint in order to normalize the output JSONstreams if needed.

**Example:**
```
python nck/entrypoint/py --normalize-keys true <reader> [...] <writer>
```
=> if `<reader>` initially returns a JSONStream, the stream will be changed in a NormalizedJSONStream

### DIFF
--- a/nck/entrypoint.py
+++ b/nck/entrypoint.py
@@ -19,10 +19,18 @@ import click
 
 from nck.writers import writers, Writer
 from nck.readers import readers, Reader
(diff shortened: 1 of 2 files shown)


gabrielleberanger (approved):
Merci beaucoup ! 👍

PR #105, opened 2021-02-17
fix: update mytarget

tom-grivaud:
### Description
- [x] Move the date check function to make it work
- [x] In rare cases the API indicates that it can't perform the request now, added a loop to retry in case the result is not fitting our expectations.

### Commits

- [x] Move the check date function to make it work.
- [x] Add a loop while the response doesn't contain the elements required.

### DIFF
--- a/nck/readers/mytarget_reader.py
+++ b/nck/readers/mytarget_reader.py
@@ -186,7 +186,7 @@ class MyTargetReader(Reader):
                 content_by_date.append(new_line)
         yield from content_by_date
 
-    @retry(wait=wait_exponential(multiplier=1, min=1, max=600), stop=stop_after_delay(600),)
+    @retry(wait=wait_exponential(multiplier=1, min=1, max=600), stop=stop_after_delay(600))
     def __get_response(self, name_content: str, offset=0) -> Dict[str, Any]:
         """This function makes a request to the api after building eveything necessary to get the
         desired results for a specific need which is defined by name_content.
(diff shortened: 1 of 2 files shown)


nck/readers/mytarget_reader.py
```diff
@@ -137,12 +136,21 @@ def __get_all_results(self, name_content: str, offset=0) -> List[Dict[str, str]]
...
+            for offset in range_offset:
+                resp = self.__get_response(name_content, offset=offset)
+                count = 0
+                while 'items' not in resp.keys() and count < 10:
+                    resp = self.__get_response(name_content, offset=offset)
+                    count += 1
+                if 'items' in resp.keys():
+                    elements.append(resp['items'])
+                else:
+                    logging.warning('Incorrect response from the API' + str(resp))
```
bibimorlet, line 153: Don't you want to use something like the **"retry"** decorator used in:
-dcm_client
-facebook_reader
-dbm_reader
...
?

bibimorlet: for example [here](https://github.com/artefactory/nautilus-connectors-kit/blob/630ba9ce882391c2d722e535e6b92c8fb4c2bfa7/nck/readers/dbm_reader.py#L153-L163) in dbm_reader

tom-grivaud:
@bibimorlet Let me know if it's good to go for you, works perfectly on my side.

bibimorlet (approved):
👍

PR #108, opened 2021-02-18
discovery Log enabled in debug mode

### LINKED ISSUE #100: As a user, I can see the HTTP request content in the logs for Google Services
WHY
Currently, we use the Discovery SDK, which hides a lot of complexity. However, for troubleshooting it could be useful to see the HTTP endpoint and the query body when we use NCK in a debug mode. This is always asked when we contact the support.

HOW
Add a log statement (debug level) before executing the request with the SDK. This log should contain as much information as possible about the request: endpoint, auth method, payload ...

louisRDSC:
### Issue

- [ ] My PR resolves #100 

### Description

- [ ] created 2 decorators allowing https details into logs (1 is for the reader functions the other one is for the __init__)
- [ ] Used these decorators in DV360_reader
- [ ] I added the possibility to set the Log level with an env variable

### DIFF
--- a/nck/config.py
+++ b/nck/config.py
@@ -20,6 +20,8 @@ import sys
 import os
 
 LEVEL = logging.INFO
+if "LOGGING_LEVEL" in os.environ:
+    LEVEL = os.environ["LOGGING_LEVEL"]
 FORMAT = "%(asctime)s - (%(name)s) - %(levelname)s - %(message)s"
 HANDLERS = [logging.StreamHandler(sys.stdout)]
 
--- a/nck/config.py
+++ b/nck/config.py
@@ -19,9 +19,7 @@ import logging
 import sys
 import os
 
-LEVEL = logging.INFO
-if "LOGGING_LEVEL" in os.environ:
-    LEVEL = os.environ["LOGGING_LEVEL"]
+LEVEL = os.environ.get("LOGGING_LEVEL", logging.INFO)
 FORMAT = "%(asctime)s - (%(name)s) - %(levelname)s - %(message)s"
 HANDLERS = [logging.StreamHandler(sys.stdout)]
 
--- a/nck/utils/stdout_to_log.py
+++ b/nck/utils/stdout_to_log.py
@@ -1,0 +1,43 @@
+import logging
+import sys
+
+
+class STDoutToLog:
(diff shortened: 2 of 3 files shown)


nck/config.py
```diff
@@ -20,6 +20,8 @@
 import os
 
 LEVEL = logging.INFO
+if "LOGGING_LEVEL" in os.environ:
+    LEVEL = os.environ["LOGGING_LEVEL"]
```
benoitgoujon, line 24: One way to handle default values for `dict` is to use the `.get()` method. What do you think about this?
```suggestion
LEVEL = os.environ.get("LOGGING_LEVEL", logging.INFO)
```

nck/utils/stdout_to_log.py
```diff
@@ -0,0 +1,68 @@
...
+            self.content = []
+
+    def flush(self):
+        pass
+
+
+def http_log(logger_name, level=logging.DEBUG):
+    def decorator(func):
+        def wrapper(*args, **kwargs):
+            httplib2.debuglevel = 4
```
benoitgoujon, line 37: Is there a constant with an explicit name that hides this magic number 4. Would be great for readability.

benoitgoujon (commented):
Thank you for this contribution 👍 

I think you need to install the pre-commit hooks to follow the code quality guidelines. Lines have been shortened for ex, so use our pre-commit hooks to avoid this.

nck/config.py
louisRDSC: yes you're right it is shorter and clearer.

nck/utils/stdout_to_log.py
louisRDSC: No there isn't. But after further checking it seems that any value greater than 0 activate debug mode.
```suggestion
            httplib2.debuglevel = 1
```
Do you prefer it like that?

benoitgoujon: In the [documentation](https://httplib2.readthedocs.io/en/latest/libhttplib2.html#httplib2.debuglevel), they say that this is the amount of information we want to print. So, with 1, we may miss some info. Let's keep it that way 🙂

benoitgoujon (approved):

PR #110, opened 2021-02-22
fix: edit reader/writer instanciation

tom-grivaud:
### Description

- [x] Encountered an issue for mytarget tokens generation an usage due to the fact that there

### Commits

- [x] Just edit the moment when the instance is created.

IMPORTANT NOTE: There is an issue with the entrypoint which is more general and who might need a small fix. When generating the list of provided readers and writers in the entrypoint.py file instances are created in the isinstance(cmd(), Reader/Writer). This causes 3 initiations of our classes instead of one which can lead to issues or useless logs. Maybe using class variables instead of Reader/Writer objects would be better to retrieve the correct cmds. Is there a specific reason why it is how it is today or can we fix this later ? @bibimorlet @gabrielleberanger @benoitgoujon

### DIFF
--- a/nck/entrypoint.py
+++ b/nck/entrypoint.py
@@ -41,17 +41,17 @@ def build_commands(cli, available_commands):
 
 @cli.resultcallback()
 def process_command_pipeline(provided_commands, normalize_keys):
-    provided_readers = [cmd() for cmd in provided_commands if isinstance(cmd(), Reader)]
-    provided_writers = [cmd() for cmd in provided_commands if isinstance(cmd(), Writer)]
+    provided_readers = [cmd for cmd in provided_commands if isinstance(cmd(), Reader)]
+    provided_writers = [cmd for cmd in provided_commands if isinstance(cmd(), Writer)]
     _validate_provided_commands(provided_readers, provided_writers)
 
     reader = provided_readers[0]
-    for stream in reader.read():
+    for stream in reader().read():
         for writer in provided_writers:
             if normalize_keys and issubclass(stream.__class__, JSONStream):
-                writer.write(NormalizedJSONStream.create_from_stream(stream))
+                writer().write(NormalizedJSONStream.create_from_stream(stream))
             else:
-                writer.write(stream)
+                writer().write(stream)
 
 
 def _validate_provided_commands(provided_readers, provided_writers):
--- a/nck/entrypoint.py
+++ b/nck/entrypoint.py
@@ -41,17 +41,17 @@ def build_commands(cli, available_commands):
 
 @cli.resultcallback()
 def process_command_pipeline(provided_commands, normalize_keys):
-    provided_readers = [cmd for cmd in provided_commands if isinstance(cmd(), Reader)]
-    provided_writers = [cmd for cmd in provided_commands if isinstance(cmd(), Writer)]
-    _validate_provided_commands(provided_readers, provided_writers)
+    cmd_instances = [cmd() for cmd in provided_commands]
+    provided_readers = list(filter(lambda o: isinstance(o, Reader), cmd_instances))
+    provided_writers = list(filter(lambda o: isinstance(o, Writer), cmd_instances))
 
     reader = provided_readers[0]
-    for stream in reader().read():
+    for stream in reader.read():
         for writer in provided_writers:
             if normalize_keys and issubclass(stream.__class__, JSONStream):
-                writer().write(NormalizedJSONStream.create_from_stream(stream))
+                writer.write(NormalizedJSONStream.create_from_stream(stream))
             else:
-                writer().write(stream)
+                writer.write(stream)
 
 
 def _validate_provided_commands(provided_readers, provided_writers):


benoitgoujon:
Can you please open an issue @tom-grivaud describing what you said with the following info please:

- 3 object instances ok but how did you notice? Can you provide a POC (a few lines of code or a command) for reproducibility ?
- a more detailed explanation of your suggestion (code samples would be greatly appreciated). I have to admit I didn't understand your suggestion.

Then, you can link the issue with this PR. Thank you!

tom-grivaud:
> Can you please open an issue @tom-grivaud describing what you said with the following info please:
> 
> * 3 object instances ok but how did you notice? Can you provide a POC (a few lines of code or a command) for reproducibility ?
> * a more detailed explanation of your suggestion (code samples would be greatly appreciated). I have to admit I didn't understand your suggestion.
> 
> Then, you can link the issue with this PR. Thank you!

First, just to make sure we're on the same page, the changes I have made here are not directly linked to this issue.

----

Well actually each time you make a valid NCK request, no matter the reader you will see that the init logs are shown 3 times. Otherwise just adding a debug log in the init of the class will be enough to see that it will be shown.
Therefore, I don't think a POC is really necessary for this and I can't provide a command as they contain credentials so I don't think there is something to do on my side about this for now. Anyone can notice it pretty easily with his own commands.

Sure I just wanted to know more about the reasons it has be done this way to make sure changing this before opening an issue for no reason.

The issue comes from the the fact that before instantiating the first reader there are already 2 previous readers which come from the cmd() in the two list comprehensions.

Currently: 
```
def process_command_pipeline(provided_commands, normalize_keys):
    provided_readers = [cmd for cmd in provided_commands if isinstance(cmd(), Reader)]
    provided_writers = [cmd for cmd in provided_commands if isinstance(cmd(), Writer)]
    _validate_provided_commands(provided_readers, provided_writers)

    reader = provided_readers[0]
    for stream in reader().read():
        for writer in provided_writers:
            if normalize_keys and issubclass(stream.__class__, JSONStream):
                writer().write(NormalizedJSONStream.create_from_stream(stream))
            else:
                writer().write(stream)
```

I don't know if there is a better way to do it but the idea would be to use cmd instead of cmd() and replace Reader/Writer.

```
def process_command_pipeline(provided_commands, normalize_keys):
    provided_readers = [cmd for cmd in provided_commands if isinstance(cmd, other)]
    provided_writers = [cmd for cmd in provided_commands if isinstance(cmd, other)]
    _validate_provided_commands(provided_readers, provided_writers)

    reader = provided_readers[0]
    for stream in reader().read():
        for writer in provided_writers:
            if normalize_keys and issubclass(stream.__class__, JSONStream):
                writer().write(NormalizedJSONStream.create_from_stream(stream))
            else:
                writer().write(stream)
```
It's likely that there are smarter ways to to this, just pretty sure the triple init comes from using cmd() twice before init the reader after.

benoitgoujon:
Ok thank you for your reply. Indeed I had noticed that before but I thought it was an issue with the loggers as we had different loggers instantiated everywhere at this time. But ok for this part.

And for the second part, I see your point, this is a problem with the current architecture and the use of Click in my opinion. Do you think that, once we will get rid of Click, the issue will still be there? We will have to refactor this part anyway, right?

tom-grivaud:
> Ok thank you for your reply. Indeed I had noticed that before but I thought it was an issue with the loggers as we had different loggers instantiated everywhere at this time. But ok for this part.
> 
> And for the second part, I see your point, this is a problem with the current architecture and the use of Click in my opinion. Do you think that, once we will get rid of Click, the issue will still be there? We will have to refactor this part anyway, right?

Hmmm I don't think getting rid of click will change anything. Getting rid of it won't change the fact that an instantiation is done by using cmd() in the list comprehensions. I don't have a solution for now, IMO we need a condition using cmd instead of cmd() that is similar to the one we currently use.

nck/entrypoint.py
```diff
@@ -41,17 +41,17 @@ def build_commands(cli, available_commands):
 
 @cli.resultcallback()
 def process_command_pipeline(provided_commands, normalize_keys):
```
gabrielleberanger, line 43: Thank you for spotting the issue @tom-grivaud. Actually this is me who implemented this specific behavior when refactoring the code (thinking that it would make `nck/entrypoint.py` more readable), but I didn't think of this downside. We could change it for the below snippet (which is actually what was implemented before):

```
@cli.resultcallback()
def process_command_pipeline(provided_commands, normalize_keys):
    cmd_instances = [cmd() for cmd in provided_commands]
    provided_readers = list(filter(lambda o: isinstance(o, Reader), cmd_instances))	
    provided_writers = list(filter(lambda o: isinstance(o, Writer), cmd_instances))
    
    reader = provided_readers[0]
    for stream in reader.read():
        for writer in provided_writers:
            if normalize_keys and issubclass(stream.__class__, JSONStream):
                writer.write(NormalizedJSONStream.create_from_stream(stream))
            else:
                writer.write(stream)
```

tom-grivaud: Yep that works better indeed. Can I push this snippet on this PR?

bibimorlet: yes

bibimorlet:
resolves #111

gabrielleberanger (approved):

bibimorlet (approved):

PR #113, opened 2021-02-24
Add data budget

tom-grivaud:
### Description

The goal here was to add the status and date of campaigns for budget reports.

Also make start/end dates uniform whether they are retrieved from a date range or not. (Click seemed to have no date option, datetime only and the util function send a date)

### Commits

- [x] Add helper to handle the new case

- [x] Create a new function to merge the data

- [x] Adapt the date type because start date end date retrieved from the get_date_start_and_date_stop_from_date_range function and the values retrieved using click.

### DIFF
--- a/nck/readers/mytarget_reader.py
+++ b/nck/readers/mytarget_reader.py
@@ -78,9 +78,11 @@ class MyTargetReader(Reader):
                 complete_daily_content = self.map_campaign_name_to_daily_stat(dict_stat, dict_camp, dict_banner)
                 yield JSONStream("mytarget_performance_", self.split_content_by_date(complete_daily_content))
             if self.request_type == "budget":
+                res_dates = self.__get_all_results("get_campaign_dates")
                 res_budgets = self.__get_all_results("get_campaign_budgets")
 
-                yield JSONStream("mytarget_budget_", self.__yield_from_list(res_budgets))
+                budget_with_dates = self.map_budget_to_date_range(res_dates, res_budgets)
+                yield JSONStream("mytarget_budget_", self.__yield_from_list(budget_with_dates))
 
     def __check_date_input_validity(self) -> bool:
         """The goal of this function is to check the validity of the date input parameters before retrieving the data."""
@@ -158,6 +160,18 @@ class MyTargetReader(Reader):
         for unused_ban_id in unused_banners:
             dict_banner.pop(unused_ban_id)
         return dict_banner
+
+    def map_budget_to_date_range(
+        self, dates: Dict[str, str], budgets: List[Dict[str, str]]
+    ) -> List[Dict[str, str]]:
+        result = []
+        dates_dict = self.__transform_list_dict_to_dict(dates)
+        for budget in budgets:
+            budget["date_start"] = dates_dict[budget["id"]]["date_start"]
+            budget["date_end"] = dates_dict[budget["id"]]["date_end"]
+            budget["status"] = dates_dict[budget["id"]]["status"]
+            result.append(budget)
+        return result
 
     def split_content_by_date(self, content: List[Dict[str, Any]]):
         """The goal of this function is to create a line for each date from the date range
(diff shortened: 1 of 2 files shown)


nck/readers/mytarget_reader.py
```diff
@@ -100,6 +99,12 @@ def __check_end_posterior_to_start(self, start_date: datetime, end_date: datetim
         else:
             return True
 
+    def __get_valid_start_date_end_date(self, date_range: str, start_date: datetime, end_date: datetime):
```
benoitgoujon, line 102: (Optional) Could we use what @ali-artefact introduced in its PR #95 to validate the input?

benoitgoujon (approved):
LGTM 👍

nck/readers/mytarget_reader.py
tom-grivaud: We're actually already using it. I placed it as the first line of our init as if the params provided are not correct we do not go further. Is it ok for you ?

Here it is: 

```
def __init__(self, client_id, client_secret, refresh_token, request_type, date_range, start_date, end_date, **kwargs):
        check_date_range_definition_conformity(start_date, end_date, date_range)
```

tom-grivaud: Speaking about this, what do you guys think about addin these function to utils about dates with the tests corresponding to it soon ? 

Here is what I am talking about:

```
def __check_date_input_validity(self) -> bool:
        """The goal of this function is to check the validity of the date input parameters before retrieving the data."""
        return self.__check_end_posterior_to_start(self.start_date, self.end_date) and self.__check_date_not_in_future(
            self.end_date
        )

def __check_date_not_in_future(self, end_date: datetime) -> bool:
        if end_date <= date.today():
            return True
        else:
            raise ValueError(f"The end date {end_date} is posterior to current date {date.today()}")

def __check_end_posterior_to_start(self, start_date: datetime, end_date: datetime) -> bool:
        if start_date > end_date:
            raise ValueError(f"The start date {start_date} is posterior to end date {end_date}")
        else:
            return True
```

The idea is to check if the date make sense here and not just consider their types etc...

benoitgoujon: Yes my bad, read a bit too fast. It's ok but maybe it would be better to group all the logic around dates together in the `__init__`. So, a quick win would bin to check first and then put your line `self.start_date ...`. I think it would help readability. But again, it's optional 🙂

benoitgoujon: Sure! Go ahead 🚀  You can follow the same design with functions that raise custom exceptions.

tom-grivaud: Totally get you, I thought about doing this at first (like I did before). But after thinking about it I think it's cleaner this way. Let me shows you what would have to be done so that it works the same way.

```suggestion
class MyTargetReader(Reader):
    def __init__(self, client_id, client_secret, refresh_token, request_type, date_range, start_date, end_date, **kwargs):
        check_date_range_definition_conformity(start_date, end_date, date_range)
        start_date, end_date = self.__get_valid_start_date_end_date(date_range, start_date, end_date)
        self.client_id = client_id
        self.client_secret = client_secret
        self.agency_client_token = {"refresh_token": refresh_token}
        self.request_type = request_type
        self.date_range = date_range
        self.start_date = start_date
        self.start_date = end_date
        self.date_format = kwargs.get("date_format")
        self.date_are_valid = self.__check_date_input_validity()
        self.__retrieve_and_set_token()
```

tom-grivaud: Suggestion was not the good move here but basically thought it was better to directly set start and end date from the result of the function to get start/end dates. Le me know what's better in your opinion and I'll do the change.

benoitgoujon: Yes that's great!

tom-grivaud: Alright, that's done. Regarding the function about the logical sense of dates provided I will create and test reusable functions later in another PR.

Issue #96, opened 2021-02-08
As a User, I can launch a request on any reader using a pre-defined date_range option [Continuation]


gabrielleberanger:
WHY
Currently, only specific readers accept a `--date_range` parameter (it depends of the source API, that offers it or not).

Such a pre-defined `--date_range` option can be very convenient, in particular when conducting tests. In this case, you just have to specify `--date-range PREVIOUS_WEEK` for instance, instead of providing a specific `--start-date` and `--end-date` for your request.

HOW
Make the `--date_range` option available to all readers, including when it is not initially provided by the source API.

**Case 1 - If the `--date_range` option is provided by the source API**
Noting to do

**Case 2 - If the `--date_range` option is not provided by the source API**
Make the the option available, using the `get_date_start_and_date_stop_from_date_range()` function (available [here](https://github.com/artefactory/nautilus-connectors-kit/blob/dev/nck/utils/date_handler.py)). Here is the status of the implementation on readers falling within this scope:
- DCM reader - To be done
- SA360 - To be done
- MyTarget - To be done
- Adobe 2.0 - Implemented
- Search Console - To be done
- Twitter Reader - To be done
- The Trade Desk - Implemented

Also see issue #67 for a description of how each reader handles date parameters.

PR #114, opened 2021-02-24
Predefined date range option

pol-defont-reaulx:
### Issue

- My PR addresses the following Issue: #96 

### Description

- add a build_date_range function in `nck/utils/date_handler.py` to manage optional date range when needed in a reader
- modify `nck/readers/adobe_reader_2_0.py` to use this new function instead of a built-in one
- add tests for this new function and adapt an existing test for adobe reader 2.0

### Tests

My PR adds the following unit tests:
- `nck/utils/test_date_handler.py` create `test_build_date_range_without_dates` to test the build_data_range function only with a date range
- `nck/utils/test_date_handler.py` create `test_build_date_range_with_dates` to test the build_data_range function only with dates
- `nck/readers/test_adobe_reader_2_0.py:` modify `test_build_date_range` to `test_format_date_range` after changing the function

### DIFF
--- a/nck/readers/search_console_reader.py
+++ b/nck/readers/search_console_reader.py
@@ -80,8 +80,6 @@ class SearchConsoleReader(Reader):
         self.dimensions = list(dimensions)
         self.site_url = site_url
         self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
-        self.start_date = datetime.strftime(self.start_date, DATEFORMAT)
-        self.end_date = datetime.strftime(self.check_end_date(self.end_date), DATEFORMAT)
         self.with_date_column = date_column
         self.row_limit = row_limit
 
@@ -109,21 +107,15 @@ class SearchConsoleReader(Reader):
 
     @staticmethod
     def check_end_date(end_date):
-        # adapt the class to be able to compare end_date with MAX_END_DATE (datetime)
-        if isinstance(end_date, datetime):
-            max_end_date = MAX_END_DATE
-        else:
-            max_end_date = MAX_END_DATE.date()
-
-        if end_date > max_end_date:
+        if end_date > MAX_END_DATE:
             logger.warning(f"The most recent date you can request is {datetime.strftime(MAX_END_DATE, DATEFORMAT)}")
         return end_date
 
     def build_query(self):
 
         query = {
-            "startDate": self.start_date,
-            "endDate": self.end_date,
+            "startDate": datetime.strftime(self.start_date, DATEFORMAT),
+            "endDate": datetime.strftime(self.check_end_date(self.end_date), DATEFORMAT),
             "dimensions": self.dimensions,
             "startRow": self.start_row,
             "rowLimit": self.row_limit,
@@ -156,7 +148,7 @@ class SearchConsoleReader(Reader):
 
             for dimension, key in zip(self.dimensions, keys):
                 if self.with_date_column:
-                    record["date"] = self.start_date
+                    record["date"] = datetime.strftime(self.start_date, DATEFORMAT)
                 record[dimension] = key
 
             for metric in metric_names:
--- a/nck/readers/search_console_reader.py
+++ b/nck/readers/search_console_reader.py
@@ -28,6 +28,7 @@ from nck.commands.command import processor
 from nck.readers.reader import Reader
 from nck.streams.normalized_json_stream import NormalizedJSONStream
 from nck.utils.args import extract_args
+from nck.utils.date_handler import DEFAULT_DATE_RANGE_FUNCTIONS, build_date_range
 from nck.utils.retry import retry
 
 
@@ -42,6 +43,11 @@ from nck.utils.retry import retry
 @click.option("--search-console-end-date", type=click.DateTime(), default=None)
 @click.option("--search-console-date-column", type=click.BOOL, default=False)
 @click.option("--search-console-row-limit", type=click.INT, default=25000)
+@click.option(
+    "--search-console-date-range",
+    type=click.Choice(DEFAULT_DATE_RANGE_FUNCTIONS.keys()),
+    help=f"One of the available NCK default date ranges: {DEFAULT_DATE_RANGE_FUNCTIONS.keys()}",
+)
 @processor()
 def search_console(**params):
     return SearchConsoleReader(**extract_args("search_console_", params))
@@ -65,6 +71,7 @@ class SearchConsoleReader(Reader):
         end_date,
         date_column,
         row_limit,
+        date_range,
     ):
         self.client_id = client_id
         self.client_secret = client_secret
@@ -72,8 +79,9 @@ class SearchConsoleReader(Reader):
         self.refresh_token = refresh_token
         self.dimensions = list(dimensions)
         self.site_url = site_url
-        self.start_date = datetime.strftime(start_date, DATEFORMAT)
-        self.end_date = datetime.strftime(self.check_end_date(end_date), DATEFORMAT)
+        self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
+        self.start_date = datetime.strftime(self.start_date, DATEFORMAT)
+        self.end_date = datetime.strftime(self.check_end_date(self.end_date), DATEFORMAT)
         self.with_date_column = date_column
         self.row_limit = row_limit
 
@@ -101,7 +109,13 @@ class SearchConsoleReader(Reader):
 
     @staticmethod
     def check_end_date(end_date):
-        if end_date > MAX_END_DATE:
+        # adapt the class to be able to compare end_date with MAX_END_DATE (datetime)
+        if isinstance(end_date, datetime):
+            max_end_date = MAX_END_DATE
+        else:
+            max_end_date = MAX_END_DATE.date()
+
+        if end_date > max_end_date:
             logger.warning(f"The most recent date you can request is {datetime.strftime(MAX_END_DATE, DATEFORMAT)}")
         return end_date
 
--- a/nck/readers/ttd_reader.py
+++ b/nck/readers/ttd_reader.py
@@ -12,7 +12,6 @@
 
 from nck.config import logger
 import click
-from click import ClickException
 import requests
 from datetime import timedelta
 from tenacity import retry, wait_exponential, stop_after_delay
@@ -78,12 +77,6 @@ class TheTradeDeskReader(Reader):
         self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
         # Report end date is exclusive: to become inclusive, it should be incremented by 1 day
         self.end_date = self.end_date + timedelta(days=1)
-
-        self._validate_dates()
-
-    def _validate_dates(self):
-        if self.end_date - timedelta(days=1) < self.start_date:
-            raise ClickException("Report end date should be equal or ulterior to report start date.")
 
     def _get_access_token(self):
         url = f"{API_HOST}/authentication"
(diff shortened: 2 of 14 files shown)


benoitgoujon:
What was the exact scope of the issue #96 ? I thought it was for all readers and not only Adobe v2. Maybe we could add the list to the issue directly to make it clear once and for all. @gabrielleberanger what do you think?

tom-grivaud:
> What was the exact scope of the issue #96 ? I thought it was for all readers and not only Adobe v2. Maybe we could add the list to the issue directly to make it clear once and for all. @gabrielleberanger what do you think?

Actually @pol-defont-reaulx asked us if it was a good idea to add his new feature to all readers. I said that IMO it wasn't a good idea as it can lead to errors if the function is not integrated at the right place for example. This happened recently and some readers were broken because a new date function was placed in many readers but not at the right place. Therefore, I think it's better that users who have credentials to test the use of the new feature add it to readers step by step. Adding it in a reader without testing it with creadentials is not a good idea to me.

pol-defont-reaulx:
I did it for the adobe reader because the issue was resolved for this particular reader, thus when I created the new function I had to adapt the reader to still support the option. As @tom-grivaud said, we discussed about integrating the solution to all readers but at least now the option can be implemented easily if needed, using this function.

gabrielleberanger:
Hi guys,

I believe there was quite a misunderstanding on issue #96.

The initial purpose of this issue was to replicate the solution that had already been implemented by @benoitgoujon (i.e. the `get_date_start_and_date_stop_from_date_range()` function, used in Adobe 2.0 and MyTarget readers) to other readers. The solution could be replicated as it is, with no further improvements.

I believe that @pol-defont-reaulx did not understand it this way, as he enhanced the existing solution with a new layer: he created a `build_date_range()` function, combining `check_date_range_definition_conformity()` and `get_date_start_and_date_stop_from_date_range()`.

What we have to settle now is: does this new layer respond to a real need?

This is true that there was a flaw in the previous version of Adobe Reader 2.0: the `build_date_range()` function (not the one implemented by @pol-defont-reaulx, but the one that was there before) reproduced the checks implemented in the `check_date_range_definition_conformity()` function, which did not make sense. So in this case, the new layer implemented by @pol-defont-reaulx has the advantage of making things clearer.

gabrielleberanger:
There is one thing I would like to clarify: what was the scope of the solution initially implemented by @benoitgoujon?
- **Option 1**: be implemented on all readers?
- **Option 2**: be implemented only on the readers whose API did not offer a pre-defined date range option?

Option 1 might be problematic as we might have a conflict between date range options provided by NCK and date range options provided by the source API, so I would tend for option 2. In this case, we would need to use:
- **When the API does not provide a date range option**: the new `build_date_range()` function defined by Pol
- **When the API provides a date range option**: the `check_date_range_definition_conformity()` and `get_date_start_and_date_stop_from_date_range()` functions separately (the new `build_date_range()` function does not work in this case, as the date range value handled by source APIs might not match the date range values handled by NCK)

Would it make sense to do so?

benoitgoujon:
The function `get_date_start_and_date_stop_from_date_range` was indeed intended for readers whose API does not offer date ranges by default. So, I would vote for option 2.

But, when the API provides a date range option, I think we don't need to use `get_date_start_and_date_stop_from_date_range`. We just need to use `check_date_range_definition_conformity` but this is out of the scope of this issue and this PR to me.

So, what we need to finish is the list of the readers that don't have the date range option and we need to provide the necessary credentials to @pol-defont-reaulx so that he can test everything works as expected.

Here all the readers that are in the scope of the original issue. This list is based on the audit done in #67 and on the readers we focus on at the moment:

- DCM reader
- SA360
- MyTarget
- Adobe V2
- SearchConsoleReader
- twitter_ready
- the Trade Desk

Is that ok for you?

gabrielleberanger:
Perfectly aligned 👍 
Thank you @benoitgoujon for the clarification!
I will add this list to the description of the issue.

@pol-defont-reaulx I can provide you with all credentials, except for MyTarget.

pol-defont-reaulx:
@gabrielleberanger yes please, I will be able to integrate the date range option to all of these readers

benoitgoujon (approved):
LGTM 👍

nck/readers/search_console_reader.py
```diff
@@ -65,15 +71,17 @@ def __init__(
...
         self.client_secret = client_secret
         self.access_token = access_token
         self.refresh_token = refresh_token
         self.dimensions = list(dimensions)
         self.site_url = site_url
-        self.start_date = datetime.strftime(start_date, DATEFORMAT)
-        self.end_date = datetime.strftime(self.check_end_date(end_date), DATEFORMAT)
+        self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
+        self.start_date = datetime.strftime(self.start_date, DATEFORMAT)
+        self.end_date = datetime.strftime(self.check_end_date(self.end_date), DATEFORMAT)
```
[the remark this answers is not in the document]
pol-defont-reaulx: I didn't change these lines I just adapted them to use "self" instead of the arguments, but I can change the code if you want to keep them as datetime

pol-defont-reaulx:
There is still a little bug on the twitter reader, I'm working on it.
I also fixed the Black's pre-commit version as discussed with @benoitgoujon and @bibimorlet because pre-commit was using its own version (and not the one in the requirements) and was getting the last version of Black for every new contributor. This was causing inconsistence between people for the code format. I decided to keep the version 19.10b0 that was in the requirements.

nck/readers/search_console_reader.py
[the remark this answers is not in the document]
benoitgoujon: Ok, it's up to you. In my opinion, manipulating native date objects until we **really** need to convert them makes a lot of things easier, especially when we have to make operations on them. Here, we don't do much with the dates, so the way it is implemented is acceptable. This is why it was a "nitpick" 🙂

pol-defont-reaulx:
Hi all,
The PR is finished for every readers mentioned in #96 and I made a bunch of little fixes.
@tom-grivaud @gabrielleberanger tell me if it's ok for you

nck/readers/search_console_reader.py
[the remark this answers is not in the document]
pol-defont-reaulx: I agree and I changed it to make it cleaner

nck/readers/ttd_reader.py
```diff
@@ -72,26 +67,17 @@ def the_trade_desk(**kwargs):
...
         self.report_template_name = report_template_name
         self.report_schedule_name = report_schedule_name
-        self.start_date = start_date
+        self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
         # Report end date is exclusive: to become inclusive, it should be incremented by 1 day
-        self.end_date = end_date + timedelta(days=1)
-        self.normalize_stream = normalize_stream
+        self.end_date = self.end_date + timedelta(days=1)
 
         self._validate_dates()
```
gabrielleberanger, line 82: I think that we don't need the `self._validate_dates()` check anymore (as we are already checking dates conformity in your  newly implemented `build_date_range()` function ) 😊

nck/readers/twitter_reader.py
```diff
@@ -171,8 +179,8 @@ def __init__(
         # General inputs
         self.report_type = report_type
         self.entity = entity
-        self.start_date = start_date
-        self.end_date = end_date + timedelta(days=1)
+        self.start_date, self.end_date = build_date_range(start_date, end_date, date_range)
```
gabrielleberanger, line 182: Same for Twitter reader, we can probably remove the `self.validate_dates()` check below 😊

gabrielleberanger (commented):
Two little comments on checks that have become unnecessary, otherwise, looks good to me! Thank you Pol! 😊

gabrielleberanger (approved):

nck/readers/twitter_reader.py
pol-defont-reaulx: Done 😄

nck/utils/date_handler.py
```diff
@@ -63,3 +63,24 @@ def get_date_start_and_date_stop_from_date_range(date_range: str) -> Tuple[date,
...
+        date_range (str): One of the default date ranges that exist
+
+    Returns:
+        Tuple[datetime, datetime]: date start and date stop that match the date range
+    """
+    check_date_range_definition_conformity(start_date, end_date, date_range)
+
+    if date_range is not None:
+        start_date, end_date = get_date_start_and_date_stop_from_date_range(date_range)
+        start_date = datetime.combine(start_date, datetime.min.time())
```
tom-grivaud, line 83: Your solution is great, i'm just looking at the details here since it's the only thing new is the convertion from date to datetime. Wouldn't it be more explicit to do something like this: 
```suggestion
        start_date = datetime(year=start_date.year, month=start_date.month, day=start_date.day)
```
Ok it takes a bit less characters to write your option but it is also less explicit to me. I may be wrong though, maybe someone more experienced like @benoitgoujon or @gabrielleberanger have a better vision of both solutions or even come with another one.

nck/utils/date_handler.py
```diff
@@ -63,3 +63,24 @@ def get_date_start_and_date_stop_from_date_range(date_range: str) -> Tuple[date,
     """
     current_date = date.today()
     return DEFAULT_DATE_RANGE_FUNCTIONS[date_range](current_date)
+
+
+def build_date_range(start_date, end_date, date_range):
```
tom-grivaud, line 68: I'm the first one to add these when it is not really necessary so you can keep it the way it is. Nevertheless, I think the meaning of the variable is obvious while types may not be (date or datetime, that is the question). 
```suggestion
def build_date_range(start_date: date, end_date: date, date_range: str) -> Tuple[datetime, datetime]:
```
By doing so you could therefore remove the content of your docstring which may not really bring value here once the types have been specified using type hinting like I suggested. Again, your work is fine, this is just a simple suggestion you can keep it like it is now!

tom-grivaud (approved):
Good job Pol and thanks for taking from your time to enhance NCK ! It actually works for me the way it is but I still made some suggestions which I hope make sense. You can still merge but if you think these suggestions are interesting then you can do them before merging!

nck/utils/date_handler.py
tom-grivaud: Here the goal is really to delete the docstring which doesn't bring value to me except the types of arguments but you can specify them in an even more explicit way thanks to type hinting.

nck/utils/date_handler.py
benoitgoujon: I agree with you @tom-grivaud. The code works well and uses the properties of the standard library elegantly but your suggestion is indeed a bit easier to understand at first sight. Btw, we don't need to repeat ourselves with args, we can just write:

```suggestion
        start_date = datetime(start_date.year, start_date.month, start_date.day)
```

nck/utils/date_handler.py
```diff
@@ -63,3 +63,24 @@ def get_date_start_and_date_stop_from_date_range(date_range: str) -> Tuple[date,
...
+    """Returns date start and date end adapted if there is a date range.
+
+    Args:
+        start_date (date): start date
```
tom-grivaud, line 72: By docstring which doesn't bring value I mean this btw. Your naming is explicit enough as the variables are quite simple here, only the date is interesting but can be mentioned using type hinting.

nck/utils/date_handler.py
pol-defont-reaulx: Indeed, it's cleaner that way

PR #115, opened 2021-02-25
Fix: add missing command validation

gabrielleberanger:
Adding validation of provided readers and writers again in the `nck/entrypoint.py` module(it had been removed by mistake when fixing command instantiation earlier this week).


bibimorlet (approved):
bien vu 👍

PR #118, opened 2021-03-01
Enh: standardize field name normalization

gabrielleberanger:
:heavy_exclamation_mark: **If we merge this PR into dev, you will need to slightly update the commands you are sending to NCK. Please read carefully the below description to warn me if these changes cannot be easily implemented in the context of your project.**

### Issue

This PR addresses Issue #103.

### Description

Two methods were coexisting to normalize field names. By normalizing, I mean replacing any special character or white space by an underscore, to match destination requirements (e.g. BigQuery that only accepts field names containing letters, digits or underscores).

These 2 methods are detailed below:

- Field names could be normalized **at the click group level**, with the click option `--normalize-keys`:
```
python nck/entrypoint.py --normalize-keys true <READER_COMMAND> <WRITER_COMMAND>
```
- Field names could be normalized **at the reader level**, either explicitly or implicitly:
   - Explicitly: when a click option `--normalize-stream` is implemented (keys are normalized only when the user asks it)
   - Implicitly: when a `NormalizedJSONStream` is implemented (keys are normalized in any case)

This PR harmonizes normalization behaviors, implementing the following standard:
- Field names can be normalized at the click group level only
- All readers are returning a `JSONStream`

### How this change affects the commands you are providing to the application

This change affects the following readers:
- Confluence
- DCM
- Facebook
- Google Analytics
- Google Ads
- Google Sheets
- Google Cloud Storage
- MySQL
- Radarly
- Amazon S3
- Salesforce
- SA360
- Search Console
- The Trade Desk

If you are using any of these and want to keep your output data under the same format, you should add the `--normalize-keys` option to the command you are providing to the application. I.e.:

If currently, you are calling the application with:
```
python nck/entrypoint.py <READER_COMMAND> <WRITER_COMMAND>
```
Now, you should now call the application with:
```
python nck/entrypoint.py --normalize-keys true <READER_COMMAND> <WRITER_COMMAND>
```

And if you are querying Confluence or The Trade Desk:
You should also remove the `--normalize-stream true` option from your reader command.

I know that @bibimorlet is already using the `--normalize-keys` option at the click group level on some sources, so he should be able to replicate this configuration easily. My concerns are more about Sanofi: @tom-grivaud, would you be able easily to integrate this configuration to your command generator? If it implies too many changes from your side, we can of course discuss it together first to see the pros and cons.

### DIFF
--- a/nck/utils/text.py
+++ b/nck/utils/text.py
@@ -24,7 +24,12 @@ from itertools import islice
 
 
 def get_report_generator_from_flat_file(
-    line_iterator, delimiter=",", skip_n_first=0, skip_n_last=0, add_column=False, column_dict={},
+    line_iterator,
+    delimiter=",",
+    skip_n_first=0,
+    skip_n_last=0,
+    add_column=False,
+    column_dict={},
 ):
     """
     From the line iterator of a flat file:
@@ -82,7 +87,13 @@ def decode_if_needed(line):
 
 def parse_decoded_line(line, delimiter=",", quotechar='"'):
     line_as_file = StringIO(line)
-    reader = csv.reader(line_as_file, delimiter=delimiter, quotechar=quotechar, quoting=csv.QUOTE_ALL, skipinitialspace=True,)
+    reader = csv.reader(
+        line_as_file,
+        delimiter=delimiter,
+        quotechar=quotechar,
+        quoting=csv.QUOTE_ALL,
+        skipinitialspace=True,
+    )
     return next(reader)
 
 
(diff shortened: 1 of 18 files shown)


nck/utils/text.py
```diff
@@ -121,3 +132,7 @@ def reformat_naming_for_bq(text, char="_"):
     text = re.sub(r"[\s\W]+", char, text)
     text = re.sub(r"[" + char + "]+", char, text.strip())
     return text.lower()
+
+
+def strip_prefix(text, prefix):
+    return re.split(prefix, text)[-1]
```
bibimorlet, line 138: 👍

bibimorlet (approved):

bibimorlet:
Sounds good to me, works fine!

tom-grivaud (approved):

Issue #102, opened 2021-02-17
As A Developer, I can run integration tests that are based on mock data


benoitgoujon:
WHY
Currently, we don't have tests that verify the end-to-end workflow because we can't share credentials from real customers to every contributor. So, it might be a good idea to look for sandbox environments to run our tests.
For example, the FB marketing API [has one](https://developers.facebook.com/ads/blog/post/v2/2016/10/19/sandbox-ad-accounts/).

HOW
Use Docker containers to reproduce a real-world environment and run different commands. Verify the output is correct.

PR #120, opened 2021-03-02
fix: adapt new func names

tom-grivaud:
Just adapt the name of the function according to the recent edits. The new name was just not applied to sdf_to_nsjon...

Interrogation about a possible improvement:
Makes me think about something, what about including checks in CI to test the full pipeline ? I know it will take time but maybe it would only be necessary to merge from dev to a new "master/main" branch. I haven't thought about all the details about this kind CI checks but it could avoid us using version of NCK in which don't work... I had to made test manually and correct these to make sure the new NCK we're going to use would work, I think this kind of feature could be an interesting improvement.

If you guys agree and if you think this is doable I will write a new issue.

### DIFF
--- a/nck/utils/file_reader.py
+++ b/nck/utils/file_reader.py
@@ -32,7 +32,7 @@ def unzip(input_file, output_path):
 def sdf_to_njson_generator(path_to_file):
     csv_reader = CSVReader(csv_delimiter=",", csv_fieldnames=None)
     with open(path_to_file, "rb") as fd:
-        dict_reader = csv_reader.read_csv(fd)
+        dict_reader = csv_reader.read(fd)
         for line in dict_reader:
             yield line
 


gabrielleberanger (approved):
Thank you for spotting this Tom 👍 

Indeed, we are missing integration tests. It happened several times recently that we merged changes that were actually not functional, and we noticed it afterwards... As for the issue, don't you think that your suggestion overlaps the Issue #102 created by Benoit?

tom-grivaud:
> Thank you for spotting this Tom 👍
> 
> Indeed, we are missing integration tests. It happened several times recently that we merged changes that were actually not functional, and we noticed it afterwards... As for the issue, don't you think that your suggestion overlaps the Issue #102 created by Benoit?

Yes it is what Benoît was talking about. It wont be quick to do I guess but it is worth it. As you've said, it happened a lot recently and this kind of issue is critical for both current and new users.

Issue #70, opened 2020-12-16
As a Developer, I can see a clear distinction between a reader, a CLI, a config and a helper module for each reader source


gabrielleberanger:
WHY
Today, the application modules do not have a single purpose:
- The `nck/readers/<source>_reader.py` modules are defining both a CLI command and a reader class
- Modules under the `nck/helpers/` directory are a mix of configuration variables and helper functions

HOW
To make things clearer, implement the following architecture:
```
- nck/
-- readers/
--- <source>/
---- reader.py
---- cli.py # nck/entrypoint.py should point to these files
---- config.py
---- helper.py
```

Issue #69, opened 2020-12-16
As a Developer, I can rely on a standardized APIClient class to authenticate requests on any reader


gabrielleberanger:
WHY
Today, each reader uses its own authentication method: either a client class or a helper method.
Available client classes are available under the `nck/clients` directory (after the refacto, clients shared by several sources will appear under `nck/clients/`, and source-level clients will appear under `nck/readers/<source>/client.py`).

HOW

**Long term goal**
Create a unique `APIClient` class, that could handle the authentication process for all readers.
Today, a standardization effort has been undertaken with the creation of the `APIClient` class ; however, this class is only used by Yandex readers today. As we might not be able to standardize all authentification methods into a single one (some APIs are quite specific), we suggest to pursue the below preliminary goal as a start.

**Intermediary goal**
Group sources sharing a similar authentication process (e.g. Google sources) into a single client class.

PR #121, opened 2021-03-03
[WIP] Enh: refactoring repo organization

gabrielleberanger:
### Issue

This PR addresses Issue #70.

### Description

The purpose of this PR is to implement the following repo organization:
```
- nck/
-- readers/
--- <source>/
---- reader.py
---- cli.py
---- config.py
---- helper.py
```

For now, I implemented it for `adobe` and `adobe_2_0` readers.
Before I replicate this structure to other readers, could you please confirm that this structure seems relevant to you?
Thank you!

### DIFF
--- a/nck/readers/__init__.py
+++ b/nck/readers/__init__.py
@@ -32,14 +32,16 @@ from nck.readers.dbm_reader import dbm
 from nck.readers.dcm_reader import dcm
 from nck.readers.ga_reader import ga
 from nck.readers.search_console_reader import search_console
-from nck.readers.adobe_reader import adobe
-from nck.readers.adobe_reader_2_0 import adobe_2_0
 from nck.readers.radarly_reader import radarly
 from nck.readers.yandex_campaign_reader import yandex_campaigns
 from nck.readers.yandex_statistics_reader import yandex_statistics
 from nck.readers.gs_reader import google_sheets
 from nck.readers.confluence_reader import confluence
 from nck.readers.mytarget_reader import mytarget
+
+from nck.readers.adobe.cli import adobe
+from nck.readers.adobe_2_0.cli import adobe_2_0
+
 
 readers = [
     mysql,
--- a/nck/readers/adobe/cli.py
+++ b/nck/readers/adobe/cli.py
@@ -1,0 +1,82 @@
+# GNU Lesser General Public License v3.0 only
+# Copyright (C) 2020 Artefact
+# licence-information@artefact.com
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import click
+from nck.commands.command import processor
+from nck.readers.adobe.reader import AdobeReader
+from nck.utils.args import extract_args
+
+
+def format_key_if_needed(ctx, param, value):
+    """
+    In some cases, newlines are escaped when passed as a click.option().
+    This callback corrects this unexpected behaviour.
+    """
+    return value.replace("\\n", "\n")
+
+
+@click.command(name="read_adobe")
+@click.option(
+    "--adobe-client-id",
+    required=True,
+    help="Client ID, that you can find in your integration section on Adobe Developper Console.",
+)
+@click.option(
+    "--adobe-client-secret",
+    required=True,
+    help="Client Secret, that you can find in your integration section on Adobe Developper Console.",
+)
+@click.option(
+    "--adobe-tech-account-id",
+    required=True,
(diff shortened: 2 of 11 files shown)


bibimorlet:
What is our final decision about the <source>_client files or a standard client entrypoint like an api_client file?
For now Adobe, CM, SA360 have client handlers

gabrielleberanger:
> What is our final decision about the _client files or a standard client entrypoint like an api_client file?
> For now Adobe, CM, SA360 have client handlers

I believe that our target is still to have a unique client class.... I see that we have a generic `APIClient` class, which is actually not generic (?) as it is used on Yandex readers only.

In the meantime, should I add a `client.py` file under each `nck/readers/<source>/` (for Adobe, CM, SA360 and Yandex)?

bibimorlet:
Yes meanwhile, I think we can add the client files. The purpose is in fact to create a generic client entrypoint like api_client that is only used by Yandex fr now. This was a feature developped by @benoitgoujon.

gabrielleberanger:
> Yes meanwhile, I think we can add the client files. The purpose is in fact to create a generic client entrypoint like api_client that is only used by Yandex fr now. This was a feature developped by @benoitgoujon.

I see ! So I should probably keep `api_client.py` in a high-level directory (to show that it is generic and not linked to a specific reader).

benoitgoujon:
Thank you @gabrielleberanger, this is way cleaner.

Regarding the decision about API clients. Indeed I had started something more generic for the Yandex reader, but this may not be possible for each reader and this is a long term vision. But one intermediary step could be to group clients from the same together. It would make sense, especially for Google readers. What do you think?

bibimorlet:
> Thank you @gabrielleberanger, this is way cleaner.
> 
> Regarding the decision about API clients. Indeed I had started something more generic for the Yandex reader, but this may not be possible for each reader and this is a long term vision. But one intermediary step could be to group clients from the same together. It would make sense, especially for Google readers. What do you think?

yes indeed, it won't be possible for every readers.
I suggest that this PR handles the file reorganization and we will group clients in another one

benoitgoujon:
Yes of course! Let's keep our PR short 😜

gabrielleberanger:
Thank you for your clarification @benoitgoujon ! I also had some doubts on the fact that we would be able to standardize all clients into a single one, so grouping similar ones seems more realistic as preliminary step. I will detail this on Issue #69 😊 

For now, I included `client.py` files under each `nck/readers/<source>/` directory.
And the `nck/clients/` directory has the following structure:
```
nck/clients/
- adobe_analytics/
-- client.py # Former nck/clients/adobe_client.py. This client was not moved to `nck/readers/<source>/` as it is shared by the two Adobe Analytics readers.
- api/
-- client.py # Former nck/clients/api_client.py
-- helper.py # Former nck/helpers/api_client_helper.py
- google/
-- client.py # Former nck/helpers/google_base.py
```
Does it make sense to you?
Also, with the new organization, I could completely remove our former `nck/helpers/` directory.

**Next steps**

As this PR implements a lot of changes, I would like to run test commands for each source/destination to make sure that it will not break anything in production when it will be released. Using Samsung credentials, I can test:

- **Readers**: Adobe Analytics 1.4, Adobe Analytics 2.0, Confluence, Facebook, Google Ads, Google Analytics, DBM, DCM, SA360, Search Console, The Trade Desk and Twitter
- **Writers**: Google Cloud Storage

@tom-grivaud, if you are using sources/destinations that are not in this list, could you please provide me sample commands, so that I can make sure that everything goes well for Sanofi also? Thank you! 😊 

It will probably help you "trust" my changes more, so that you won't have to review the 180 changed files one by one :stuck_out_tongue:

benoitgoujon:
Yes, would make perfect sense!

Yes, we need to be careful on that before merging 👍

gabrielleberanger:
I launched test commands on all the sources/destinations @tom-grivaud and I are using, and it works fine :ok_hand:
So I will adapt the documentation, and we'll be good for merging.

README.md
```diff
@@ -1,90 +1,65 @@
 # Nautilus Connectors Kit
 
-**NCK is a Command-Line Interface (CLI), allowing you to easily request, stream and store raw reports, from the API source to the destination of your choice.**
+**NCK is an ET(L) tool specialized in API data ingestion. It is accessible through a Command-Line Interface. The application allows you to easily extract, stream and load data (with minimum transformations), from the API source to the destination of your choice.**
```
benoitgoujon, line 3: Is it really the "L" that is optional?

```suggestion
**NCK is an E(T)L tool specialized in API data ingestion. It is accessible through a Command-Line Interface. The application allows you to easily extract, stream and load data (with minimum transformations), from the API source to the destination of your choice.**
```

docs/source/getting_started.rst
```diff
@@ -181,21 +181,44 @@ How to develop a new reader
 
 To create a new reader, you should:
```
benoitgoujon, line 182: This is not the purpose of this PR but the  "Getting Started" is oriented towards contributions and not towards our end users. I'm wondering if it is a good thing. Maybe we should put this section in a "Contributing" section and create a real "Getting started" guide with info on installation and 1 or 2 simple commands.

README.md
gabrielleberanger: Indeed, thank you for spotting this !

docs/source/getting_started.rst
gabrielleberanger: Good point! I'll create an issue for this matter.

gabrielleberanger:
I am good for merging, if you're fine with the changes 😊 

Just a precision: in the end, I chose to keep all clients in the `nck/clients` folder (I found it disturbing to have some clients in the `nck/clients` folder, and others in the `nck/readers/<source>/client.py` modules).

benoitgoujon (approved):

PR #122, opened 2021-03-04
fix date checks and format them

bibimorlet:
### Issue

- [x] No spotted issue

### Description

Date checks were inconsistent with the reader.
The reader was no longer usable with (start-end, end-date) couple parameters.

### DIFF
--- a/nck/utils/date_handler.py
+++ b/nck/utils/date_handler.py
@@ -40,18 +40,12 @@ DEFAULT_DATE_RANGE_FUNCTIONS = {
 
 
 def check_date_range_definition_conformity(start_date: date, end_date: date, date_range: str):
-    if (start_date is None and end_date is None) and date_range is None:
-        raise NoDateDefinitionException(
-            "You must at least define a couple \
-                        start-date/end-date or a date-range"
-        )
-    elif (start_date is None or end_date is None) and (start_date is not None or end_date is not None):
-        raise MissingDateDefinitionException("If you define dates, both start_date and end_date must be defined")
-    if start_date is not None and end_date is not None and date_range is not None:
-        raise InconsistentDateDefinitionException(
-            "You must define either start_date and end_date \
-            or date_range, but not both"
-        )
+    if not any([start_date, end_date, date_range]):
+        raise NoDateDefinitionException("You must at least define a couple (start-date, end-date) or a date-range")
+    elif bool(start_date) != bool(end_date):
+        raise MissingDateDefinitionException("Both start_date and end_date must be defined")
+    elif all([start_date, end_date, date_range]):
+        raise InconsistentDateDefinitionException("You must define either (start_date, end_date) or date_range but not both")
 
 
 def get_date_start_and_date_stop_from_date_range(date_range: str) -> Tuple[date, date]:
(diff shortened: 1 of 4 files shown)


nck/utils/date_handler.py
```diff
@@ -40,18 +40,12 @@ def __get_previous_month_dates(current_date: date) -> Tuple[date, date]:
...
-        raise InconsistentDateDefinitionException(
-            "You must define either start_date and end_date \
-            or date_range, but not both"
-        )
+    if not any([start_date, end_date, date_range]):
+        raise NoDateDefinitionException("You must at least define a couple (start-date, end-date) or a date-range")
+    elif bool(start_date) != bool(end_date):
+        raise MissingDateDefinitionException("Both start_date and end_date must be defined")
+    elif all([start_date, end_date, date_range]):
+        raise InconsistentDateDefinitionException("You must define either (start_date, end_date) or date_range but not both")
```
bibimorlet, line 48: Why did we define various exceptions here?
A simple DateDefinitionException is not enough?

gabrielleberanger: I agree, exception names are very close in this case, and don't allow us to make a real distinction between the different cases, so a single `DateDefinitionException` should be enough.

gabrielleberanger: Shouldn't we also throw an exception if `start_date` > `end_date`?

bibimorlet: Probably yes because in this case, the native exception thrown by the API is :
"The date range you've entered is invalid."

nck/utils/date_handler.py
```diff
@@ -40,18 +40,16 @@ def __get_previous_month_dates(current_date: date) -> Tuple[date, date]:
...
+    if bool(start_date) != bool(end_date):
+        raise MissingDateDefinitionException("Both start_date and end_date must be defined")
+    if date_range:
+        if all([start_date, end_date]):
+            raise InconsistentDateDefinitionException("You must define either (start_date, end_date) or date_range")
+    else:
+        if not any([start_date, end_date]):
+            raise NoDateDefinitionException("You must at least define a couple (start-date, end-date) or a date-range")
+        elif end_date < start_date:
+            raise InconsistentDateDefinitionException("Report end date should be equal or ulterior to report start date.")
```
gabrielleberanger, line 52: Shouldn't it rather be:
```
def check_date_range_definition_conformity(start_date: date, end_date: date, date_range: str):
    if date_range:
        if any([start_date, end_date]):
            raise DateDefinitionException("You must define either (start_date, end_date) or date_range")
    else:
        if not all([start_date, end_date]):
            raise DefinitionException("You must at least define a couple (start-date, end-date) or a date-range")
        elif end_date < start_date:
            raise DateDefinitionException("Report end date should be equal or ulterior to report start date.")
```

bibimorlet: Yes, looks better indeed. I got confused 

Otherwise I can't change the Exception names in this PR because they are used in other readers. We should create another PR t fix it.

bibimorlet: Actually I added a new generic DateDefinitionException for these checks

gabrielleberanger (approved):
LGTM 👍

PR #124, opened 2021-03-05
Storage writers factorization

### LINKED ISSUE #123: As a developper I can see standard interface and behaviour between cloud storage writers
WHY
Currently storage writers have different behaviors from each other. 

- return values of the write functions
- management of non existing buckets.

HOW
Normalize behaviors of storage writers and factorize code in a objectstorage_writer class.

louisRDSC:

### Issue

- [x] My PR resolve #123 

### Description

- [x] factorization of storage writers into objectstorage_writers.py
- [x] normalized behavior for storage writers

### DIFF
--- a/nck/writers/objectstorage_writer.py
+++ b/nck/writers/objectstorage_writer.py
@@ -33,11 +33,9 @@ class ObjectStorageWriter(Writer):
     @retry
     def write(self, stream):
         logger.info(f"Start writing file to {self._platform} ...")
-        print(self._file_name)
         self._set_valid_file_name(stream.name)
-        print(self._file_name)
         final_name = os.path.join(self._prefix, self._file_name)
-        self._create_blob(final_name, stream)
+        # self._create_blob(final_name, stream)
         logger.info(f"Wrote {final_name} file to {self._bucket_name} on  {self._platform} ...")
         uri = self._get_uri(final_name)
         logger.info(f"file can be found at {uri}")
@@ -48,10 +46,9 @@ class ObjectStorageWriter(Writer):
         list_buckets_names = [bucket.name for bucket in self._list_buckets(client)]
         try:
             assert self._bucket_name in list_buckets_names
-        except AssertionError as err:
+        except AssertionError:
             logger.exception(f"{self._bucket_name} bucket does not exist.")
             logger.exception(f"available buckets are {list_buckets_names}")
-            raise err
 
         return bucket
 
--- a/nck/writers/objectstorage_writer.py
+++ b/nck/writers/objectstorage_writer.py
@@ -1,0 +1,48 @@
+import os
+
+from nck.config import logger
+from nck.writers import Writer
+from nck.utils.retry import retry
+
+
+class ObjectStorageWriter(Writer):
+    def __init__(self, bucket, prefix=None, file_name=None, platform=None, **kwargs):
+        self._bucket_name = bucket
+        self._prefix = prefix if prefix else ""
+        self._file_name = file_name
+        self._platform = platform
+        self.bucket = self._create_bucket()
+
+    @retry
+    def write(self, stream):
+        logger.info(f"Start writing file to {self._platform} ...")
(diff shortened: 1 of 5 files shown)


nck/writers/objectstorage_writer.py
```diff
@@ -0,0 +1,78 @@
...
+        self._bucket_name = bucket_name
+        self._prefix = prefix if prefix else ""
+        self._file_name = file_name
+        self._platform = platform
+        self._bucket = self._get_bucket_if_exist()
+
+    @retry
+    def write(self, stream):
+        logger.info(f"Start writing file to {self._platform} ...")
+        print(self._file_name)
```
benoitgoujon, line 36: It is generally advised to use the `logging` package instead of `print` statements

nck/writers/objectstorage_writer.py
```diff
@@ -0,0 +1,78 @@
...
+    def _get_bucket_if_exist(self):
+        client = self._create_client()
+        bucket = self._create_bucket(client)
+        list_buckets_names = [bucket.name for bucket in self._list_buckets(client)]
+        try:
+            assert self._bucket_name in list_buckets_names
+        except AssertionError as err:
+            logger.exception(f"{self._bucket_name} bucket does not exist.")
+            logger.exception(f"available buckets are {list_buckets_names}")
+            raise err
```
benoitgoujon, line 54: Nice use of `logger.exception()`, I was not aware of this method that is indeed perfect for this use case. However, when you raise the exception, it will log something, won't it? I'm afraid the information you log before will be lost because there will be something coming afterwards. Have you tested this behaviour?

benoitgoujon (commented):
Nicely done! Do you have time to write some tests? 😬

nck/writers/objectstorage_writer.py
louisRDSC: sorry these are relicas from when i was debugging

nck/writers/objectstorage_writer.py
louisRDSC: It log what i want but the trace of the exception is print afterward, so your right, the information is quite lost in too much information. Any idea on how correct that?

louisRDSC: would this be a better way of doing it ?
```suggestion
            logger.exception(f"{self._bucket_name} bucket does not exist. available buckets are {list_buckets_names}")
            sys.exit(1)
```

benoitgoujon: Hm maybe you can change the message of the exception. Something like:

```suggestion
            err.message = f"{self._bucket_name} bucket does not exist. available buckets are {list_buckets_names}"
            raise err
```

Didn't test it, so tell me if it doesn't work.

benoitgoujon: Hm this seems like it is an old-fashioned way. Maybe you can use [with_traceback](https://docs.python.org/3/library/exceptions.html#BaseException.with_traceback) to take advantage of Python 3

tests/writers/test_s3_writer.py
```diff
@@ -0,0 +1,60 @@
...
+@mock_s3
+class S3WriterTest(TestCase):
+    @classmethod
+    @mock_s3
+    def setUpClass(cls):
+        client1 = boto3.resource("s3", region_name="us-east-1")
+        client1.create_bucket(Bucket="test")
+
+    def test_bucket_doesnt_exist(self):
+        with self.assertRaisesRegex(Exception, ".bucket does not exist."):
```
benoitgoujon, line 32: Maybe we could be more precise and test that we have the bucket in the error message like this:

```suggestion
        with self.assertRaisesRegex(Exception, "non-existing-bucket bucket does not exist."):
```

Also, I guess we can't verify that it logs the list of available buckets?

benoitgoujon (approved):
I don't know how difficult it was for you to use `moto` but the result is extra dope! LGTM 👍

Issue #87, opened 2021-01-25
As a User, I can configure/launch NCK request by passing a .json file to the application


gabrielleberanger:
WHY
Today, generating NCK CLI commands requires a lot of efforts. Every project team using the application in production had to develop a “command generator”, i.e. a tool building NCK CLI commands from a set of parameters defined on a .json file. It would be much more efficient if we could directly pass such a .json file to the application to configure/launch a request, as it would avoid making costly (and unnecessary) transformations.

HOW
Allow users to pass a .json file to the application. It can be done by setting-up:
- a second CLI that would take a single argument:
`python nck/entrypoint_json.py read_dbm --config-file <path to the .json config file>`

To be backward compatible, we could think of the following architecture:
```
- nck/
-- entrypoint/
--- entrypoint.py # Current entrypoint, allowing multiple parameters as a CLI
--- entrypoint_json.py # New entrypoint, allowing a single .json input
```

The second entrypoint will need to be completely separated from the first one and to implement a fully new architecture with a general formatter useable for every readers. The validation steps will be implemented in a separated issue.

PR #126, opened 2021-03-10
Launch NCK request by passing a .json

pol-defont-reaulx:
### Issue

- My PR addresses the following Issue: #87 

### Description

- create a folder to contain different entrypoint as nck/entrypoint/
- put entrypoint.py in the new folder.
- create launch() function in entrypoint.py to be able to launch the main functions from another entrypoint.
- create entrypoint_json.py to take a config file path as an argument or as an environment variable, parse it to get all arguments and then call launch() of entrypoint.py to launch the command described in the json file.
- I defined the following architecture for the json config file:
`{
  "--option": value,
  ...,
  "reader": {
    "name": reader_name,
    "--option": value,
     ...,
  },
  "writers": [
    {
      "name": writer_name,
      "--option": value,
      ...,
    },
    {
       ...,
    },
  ]
}`

### Documentation

I was wondering where was the best place to write the documentation according to you. Tell me.

### DIFF
--- a/nck/entrypoint/__init__.py
+++ b/nck/entrypoint/__init__.py
@@ -1,0 +1,17 @@
+# GNU Lesser General Public License v3.0 only
+# Copyright (C) 2020 Artefact
+# licence-information@artefact.com
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
--- a/nck/entrypoint/entrypoint_json.py
+++ b/nck/entrypoint/entrypoint_json.py
@@ -23,18 +23,9 @@ import nck.entrypoint.entrypoint as entrypoint
 
 @click.command()
 @click.option(
-    "--config-file",
-    help="Path of the json file used to build the command. Use this option or define an environment variable CONFIG_FILE.",
-    envvar="CONFIG_FILE",
-    type=click.Path(exists=True),
+    "--config-file", help="Path of the json file used to build the command.", required=True, type=click.Path(exists=True),
 )
 def add_json_args(config_file):
-    if config_file is None:
-        raise click.BadParameter(
-            "You must specify a config file with the option '--config-file' "
-            "or with an environment variable named CONFIG_FILE"
-        )
-
     with open(config_file) as f:
         data = json.load(f)
 
--- a/nck/entrypoint/entrypoint_json.py
+++ b/nck/entrypoint/entrypoint_json.py
@@ -1,0 +1,79 @@
+# GNU Lesser General Public License v3.0 only
+# Copyright (C) 2020 Artefact
+# licence-information@artefact.com
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+import json
+import sys
+import click
+import nck.entrypoint.entrypoint as entrypoint
+
+
+@click.command()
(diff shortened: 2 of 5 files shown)


nck/entrypoint/entrypoint_json.py
```diff
@@ -0,0 +1,79 @@
...
+
+    with open(config_file) as f:
+        data = json.load(f)
+
+    del sys.argv[1:]
+    sys.argv.extend(_parse_to_text(data))
+    print(sys.argv)
+
+
+def _parse_to_text(data):
```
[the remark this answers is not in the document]
pol-defont-reaulx: For now, we have to specify each option. I can simplify it to a list, I think it's better to avoid very long json files

nck/entrypoint/entrypoint_json.py
```diff
@@ -0,0 +1,79 @@
...
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+import json
+import sys
+import click
+import nck.entrypoint.entrypoint as entrypoint
+
+
+@click.command()
+@click.option(
```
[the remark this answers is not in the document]
pol-defont-reaulx: Yes of course

pol-defont-reaulx:
> Your new entrypoint_json.py calls the previous entrypoint.py: are we fine with this structure, or do we prefer to have 2 fully independent methods to call the application? As I understood that the long term vision would be to completely remove the CLI (to replace it by an API), aren't we more encline to a new, fully independent entrypoint?

@gabrielleberanger with the current architecture, we have to use Click to pass all arguments to the function for what I saw in the code. I think that creating a fully independent entrypoint will only result on create duplicates of entrypoint.py functions. 

I was also thinking that having two different entrypoint could be confusing for the user, especially as they are in a folder of a folder. Shouldn't it be better to create a first entrypoint in nck/ that would filter the requests between normal requests and json ones?

nck/entrypoint/__init__.py
```diff
@@ -0,0 +1,17 @@
+# GNU Lesser General Public License v3.0 only
```
benoitgoujon, line 1: Are we sure we need this file?

nck/entrypoint/entrypoint_json.py
```diff
@@ -0,0 +1,74 @@
...
+    "--config-file", help="Path of the json file used to build the command.", required=True, type=click.Path(exists=True),
+)
+def add_json_args(config_file):
+    with open(config_file) as f:
+        data = json.load(f)
+
+    del sys.argv[1:]
+    sys.argv.extend(_parse_to_text(data))
+    print(sys.argv)
```
benoitgoujon, line 34: 👀

benoitgoujon (commented):
Thanks @pol-defont-reaulx for your proposal.

As @gabrielleberanger said, I think we need to precisely define the schema of the JSON file together and then create validation functions. We can use [jsonschema](https://pypi.org/project/jsonschema/), which is really convenient for this use case.

Also, I think it's ok to duplicate code between the entrypoints for now. It will make the removal of the old one easier.

nck/entrypoint/__init__.py
gabrielleberanger: Yes, if we don't put `__init__.py` files in directories, it causes import errors when running tests with `nosetests`. I tried to remove them when doing the refacto last week, and I had to put them back because of that.

nck/entrypoint/entrypoint_json.py
pol-defont-reaulx: Oups 😅

nck/entrypoint/__init__.py
pol-defont-reaulx: The content or the file itself? For the content, I paste the content of another __init__.py file that's why

bibimorlet:
What about the Dockerfile?
In fact, the entrypoint of the Dockerfile is still:

```Dockerfile
ENTRYPOINT ["python", "nck/entrypoint.py"]
```

nck/entrypoint/__init__.py
gabrielleberanger: The content of the file is the license, we have to put it everywhere. I think Benoit was talking of the file itself :)

bibimorlet: @benoitgoujon was talking about the file itself. But as @gabrielleberanger said, the init file is visibly mandatory for nosetests

bibimorlet (changes requested):
Dockerfile

nck/entrypoint/__init__.py
pol-defont-reaulx: Yes without this file, I think the module cannot be used in other python files

pol-defont-reaulx:
@bibimorlet Indeed, I forgot to check the Dockerfile. I changed it to comply to the new name but now that we have two entrypoints I don't know how to adapt this file. Do we keep only this entrypoint in the Dockerfile?

bibimorlet:
> @bibimorlet Indeed, I forgot to check the Dockerfile. I changed it to comply to the new name but now that we have two entrypoints I don't know how to adapt this file. Do we keep only this entrypoint in the Dockerfile?

Good question? We will probably have to change it in order to handle the several ways to call NCK

nck/entrypoint/__init__.py
benoitgoujon: hm maybe we should change the way we run our tests then. Because `__init__.py` files were very Python 2. Now, we are not supposed to need them, especially if they are blank.

pol-defont-reaulx:
Hello everyone,
After a call with @gabrielleberanger and @benoitgoujon, we decided to close this PR as it asks for more complete changes in NCK's architecture. Other issues will be thought to guide new changes that will initiate a process without Click, towards a more API version of NCK. The goal will still be to answer #87 but with deeper changes.

PR #128, opened 2021-03-16
Fix/twitter api

Pierre-Alexandre35:
### Description

- [x] Here are some details about my PR, including example of usage.


nck/readers/twitter/reader.py
```diff
@@ -311,23 +311,27 @@ def get_campaign_management_report(self):
             for entity_obj in ACCOUNT_CHILD_OBJECTS[self.entity]
         ]
 
+    @retry(stop=stop_after_delay(901))
```
bibimorlet, line 314: Why this magical 901 delay?

nck/readers/twitter/reader.py
```diff
@@ -311,23 +311,27 @@ def get_campaign_management_report(self):
...
+                    card_attributes = {attr: getattr(card_fetch, attr, None) for attr in self.entity_attributes}
+                    record = {
+                        "tweet_id": tweet["tweet_id"],
+                        "card_uri": tweet["card_uri"],
+                        **card_attributes,
+                    }
+                    yield record
+            except Exception:
+                ex_type, ex, tb = sys.exc_info()
+                logger.warning(f"Failed to ingest post with error: {ex}. Traceback: {traceback.print_tb(tb)}")
```
bibimorlet, line 334: Is it necessary to get the all the logs and traceback here and not only a custom Exception?

Open discussion

nck/readers/twitter/reader.py
```diff
@@ -351,6 +354,7 @@ def get_card_fetch(self, card_uri):
 
         return CardsFetch.load(self.account, card_uris=[card_uri]).first
 
+    @retry(stop=stop_after_delay(901))
```
bibimorlet, line 357: call 911? 🤔

nck/readers/twitter/reader.py
Pierre-Alexandre35: Hi Vivien, I used 901 to avoid the rateLimit on the Twitter API ["900 requests per 15-minute interval can be made on behalf of each user"](https://developer.twitter.com/en/docs/twitter-api/v1/rate-limits). 15 * 60 = 900. I tested this solution locally for the same request and never faced a RateLimit issue anymore.

benoitgoujon: Maybe we can add a comment to explain a little bit more.

PR #132, opened 2021-03-22
poc generique stream

louisRDSC:
Here is the first implementation of a generique bytes stream. 

You can see the end to end test of it in test/end_to_end, which test writing from and to S3.


benoitgoujon:
Very simple indeed. Did you have time to try with more complex readers?

louisRDSC:
> Very simple indeed. Did you have time to try with more complex readers?

Not yet, I will try to do it by the end of the week.

PR #141, opened 2021-04-08
remove retry for s3 writer

bibimorlet:
### Issue

- With the current architecture, because of the retries in the writers, the writer put an empty file in the blob storage when an error occurs.

### Description

- Remove the retry in the writer to avoid writing an empty in the output.


louisRDSC (approved):
LGTM.

PR #147, opened 2021-04-29
Fix: google dcm reports taking more than one hour to be ready

benoitbazouin:
### Description

The goal of this PR is to update the google dcm client so that we can retrieve data when the generation of the report takes more than one hour to be completed. 
To do so:
- I increased the stop after delay parameter in the retry of the function assert_report_file_ready (3600 -> 36000)
- I added a refresh of the authorization header before the download of the report (because access tokens are valid for only one hour and are regenerated when their lifetime expires)

### DIFF
--- a/ack/clients/google_dcm/client.py
+++ b/ack/clients/google_dcm/client.py
@@ -89,7 +89,7 @@ class GoogleDCMClient:
         file_id = file["id"]
         return report_id, file_id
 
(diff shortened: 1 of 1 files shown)


tom-grivaud (approved):
Great that you made the querying time longer and thought about refreshing the token accordingly. Seems good to me !

bibimorlet (approved):

PR #153, opened 2021-06-01
modify result_generator Yandex_campaign

benoitbazouin:
### Description

- The goal of this PR is to modify the yandex_campaign reader in order to return an empty file when the result of the API is the following one: {"result": {}}. 
- For this purpose, I added a check in the result_generator before yielding the API response.

### DIFF
--- a/ack/readers/yandex_campaign/reader.py
+++ b/ack/readers/yandex_campaign/reader.py
@@ -36,7 +36,7 @@ class YandexCampaignReader(Reader):
         api_client = ApiClient(self.token, YANDEX_DIRECT_API_BASE_URL)
         request_body = self._build_request_body()
         response = api_client.execute_request(url="campaigns", body=request_body, headers={})
-        if response.json()["result"]:
+        if response.json().get("result", {}):
             yield response.json()
 
     def _build_request_body(self):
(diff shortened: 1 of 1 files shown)


ack/readers/yandex_campaign/reader.py
```diff
@@ -36,7 +36,8 @@ def result_generator(self):
         api_client = ApiClient(self.token, YANDEX_DIRECT_API_BASE_URL)
         request_body = self._build_request_body()
         response = api_client.execute_request(url="campaigns", body=request_body, headers={})
-        yield response.json()
+        if response.json()["result"]:
+            yield response.json()
```
bibimorlet: Is there always a "result" key in the JSON?

bibimorlet:
If I understand well, the purpose is to get an empty file instead of a file with '{"result":{}}', isn't it?

benoitbazouin:
> If I understand well, the purpose is to get an empty file instead of a file with '{"result":{}}', isn't it?

Yes correct

ack/readers/yandex_campaign/reader.py
benoitbazouin: I checked in the API documentation and it seemed so. I can check again to be sure but it might be a good idea to add another check just in case. What do you think ?

bibimorlet: this is probably good enough but maybe you can add:
 ```python
if response.json().get("result", {}):
```
or something like that, in order to avoid an eventual error. I don't know the best practice about it but let me know what you think! 👍

benoitbazouin: Ok thanks, indeed it may be better to use this condition instead. I'm also not sure if it is the best practice, as we might "hide" an issue instead of raising an error, but I think that in our use case it is ok to do so.

bibimorlet (approved):

tom-grivaud (approved):
LTGM 👍

PR #155, opened 2021-06-03
Fix twitter reader

louisRDSC:

### Issue

- [x] My pr adress twitterads rate limits problem

### Description

- [x] used batch card fetch, reduced by 200x the number of call to this function.

### Tests

- [x] my pr doesn't change the reader behavior so tests are still relevant

### DIFF
--- a/ack/readers/twitter/reader.py
+++ b/ack/readers/twitter/reader.py
@@ -30,6 +30,7 @@ from ack.readers.twitter.config import (
     MAX_CONCURRENT_JOBS,
     MAX_ENTITY_IDS_PER_JOB,
     REP_DATEFORMAT,
+    BATCH_SIZE
 )
 from ack.streams.json_stream import JSONStream
 from ack.utils.date_handler import build_date_range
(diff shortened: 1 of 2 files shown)


ack/readers/twitter/reader.py
```diff
@@ -316,9 +316,22 @@ def get_cards_report(self):
         Supported entities: CARD
         Documentation: https://developer.twitter.com/en/docs/ads/creatives/api-reference/
         """
+        tweets_with_uris = []
         for tweet in self.get_published_tweets_generator():
             if "card_uri" in tweet:
-                card_fetch = self.get_card_fetch(card_uri=tweet["card_uri"])
+                tweets_with_uris.append(tweet)
+
+        batch_tweets = [tweets_with_uris[i:i + 200] for i in range(0, len(tweets_with_uris), 200)]
```
bibimorlet: What is this "200"? An arbitrary batch size? The max size allowed by the Twitter API?
It would be better to put it in a constant parameter

bibimorlet (changes requested):

ack/readers/twitter/reader.py
louisRDSC: it is the maximal batch size accepted by twitter api.

bibimorlet (approved):

PR #156, opened 2021-06-03
mytarget: raise exception if http status code != 200

benoitbazouin:
### Description

- The goal of this PR is to raise an exception in the reader when we have an api response with http status != 200. 
- When this exception is raised, the code breaks without retrying and there is no file returned in the end. I did this because in our case, we got only '404' errors, with the error message "resources not found"
- The error message is logged, and the url requested too

### DIFF
--- a/ack/readers/mytarget/reader.py
+++ b/ack/readers/mytarget/reader.py
@@ -21,12 +21,13 @@ from datetime import datetime
 from typing import Any, Dict, List, Tuple
 
 import requests
-from ack.readers.mytarget.config import LIMIT_REQUEST_MYTARGET, REQUEST_CONFIG
+from ack.readers.mytarget.config import HTTP_STATUS_OK, LIMIT_REQUEST_MYTARGET, REQUEST_CONFIG
 from ack.readers.reader import Reader
 from ack.streams.json_stream import JSONStream
 from ack.utils.exceptions import MissingItemsInResponse
-from tenacity import retry, stop_after_delay, wait_exponential
+from tenacity import retry, stop_after_delay, wait_exponential, retry_if_exception_type
 from ack.utils.date_handler import build_date_range
+from ack.config import logger
 
 
 class MyTargetReader(Reader):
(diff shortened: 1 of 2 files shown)


bibimorlet (approved):

ack/readers/mytarget/reader.py
```diff
@@ -168,7 +169,11 @@ def split_content_by_date(self, content: List[Dict[str, Any]]):
                 content_by_date.append(new_line)
         yield from content_by_date
 
-    @retry(wait=wait_exponential(multiplier=1, min=1, max=600), stop=stop_after_delay(600))
+    @retry(
+        wait=wait_exponential(multiplier=1, min=1, max=600),
```
tom-grivaud, line 173: Looking at the time it took for each id, can you confirm that this 10 mins limit still makes sense ?

tom-grivaud (approved):
LGTM 👍

ack/readers/mytarget/reader.py
```diff
@@ -182,10 +187,15 @@ def __get_response(self, name_content: str, offset=0) -> Dict[str, Any]:
         """
         parameters = self.__generate_params_dict(name_content, offset=offset)
         request = self.__create_request(name_content, parameters)
-        resp = requests.get(**request).json()
-        if "items" not in resp.keys():
-            raise MissingItemsInResponse("Can't retrieve any item from this response")
-        return resp
+        resp = requests.get(**request)
+        if resp.status_code != HTTP_STATUS_OK:
```
tom-grivaud, line 191: Are you sure we never get another 2XX code here ? I'm wondering if getting a 202 for example could occur and then raise an exception while it's not really an error. Maybe need to check the API doc.

ack/readers/mytarget/reader.py
benoitbazouin: Yes, it still makes sense

ack/readers/mytarget/reader.py
benoitbazouin: The document of the API does not explain wether or not we can get another 2XX code. For our use case, we only got 200.

PR #157, opened 2021-06-22
update api version google dcm

benoitbazouin:
The Campaign Manager 360 API v3.3 is deprecated and will be sunset on June 30th, 2021. All users must migrate to a newer API version by that date.

The latest version is v3.5

I also modified the function add dimension filter. Indeed, the dimensionValues().query can only retrieve 100 results, so we have to iterate using the next page token to get all the possible values. 

Moreover, I think that it is better to raise an error when the user wants to add a dimension filter with an incorrect value. Otherwise, we still make our request to the API (with no filter added).

### DIFF
--- a/ack/clients/google_dcm/client.py
+++ b/ack/clients/google_dcm/client.py
@@ -72,6 +72,16 @@ class GoogleDCMClient:
                 "endDate": report["criteria"]["dateRange"]["endDate"],
             }
             values = self._service.dimensionValues().query(profileId=profile_id, body=request).execute()
+            current_values = values
+
+            while current_values["items"]:
+                nextPageToken = current_values["nextPageToken"]
+                current_values = (
+                    self._service.dimensionValues()
+                    .query(profileId=profile_id, body=request, pageToken=nextPageToken)
+                    .execute()
+                )
+                values["items"] += current_values["items"]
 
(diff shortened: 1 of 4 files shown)


benoitbazouin:
@tom-grivaud @bibimorlet, what do you think about these changes ?

tom-grivaud (approved):
Makes sense to me as we can get more than 100 results and that we need to go to the next page. And great for the version edit, will avoid some issues later on.

benoitbazouin:
> Makes sense to me as we can get more than 100 results and that we need to go to the next page. And great for the version edit, will avoid some issues later on.

Do you also agree that we need to raise an error when a filter value is not found ?

tom-grivaud:
> > Makes sense to me as we can get more than 100 results and that we need to go to the next page. And great for the version edit, will avoid some issues later on.
> 
> Do you also agree that we need to raise an error when a filter value is not found ?

Well if I understood it correctly, if the filter value is not found you won't get the expected data right? Therefore I prefer to raise an error instead of getting unwanted data which won't automatically raise error but could end up in wrong data in your db. Did you have any other concern or points regarding this issue ?

bibimorlet (approved):
👍

PR #158, opened 2021-07-05
Scheduled queries google dbm

benoitbazouin:
The goal of this PR is to enable the creation of scheduled reports for Google DBM reader. 

To do so, I added a function that creates the "schedule" body of the query depending on the query request type.

Moreover, I added the the possible values for the timezone codes and the frequencies in cli.py

### DIFF
--- a/ack/readers/google_dbm/cli.py
+++ b/ack/readers/google_dbm/cli.py
@@ -17,7 +17,7 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 import click
-from ack.readers.google_dbm.config import POSSIBLE_REQUEST_TYPES
+from ack.readers.google_dbm.config import POSSIBLE_REQUEST_TYPES, POSSIBLE_FREQUENCIES, POSSIBLE_TIMEZONE_CODES
 from ack.readers.google_dbm.reader import GoogleDBMReader
 from ack.utils.args import extract_args
 from ack.utils.processor import processor
@@ -33,7 +33,8 @@ from ack.utils.processor import processor
 @click.option("--dbm-request-type", type=click.Choice(POSSIBLE_REQUEST_TYPES), required=True)
 @click.option("--dbm-query-id")
 @click.option("--dbm-query-title")
-@click.option("--dbm-query-frequency", default="ONE_TIME")
+@click.option("--dbm-query-frequency", type=click.Choice(POSSIBLE_FREQUENCIES), default="ONE_TIME")
+@click.option("--dbm-query-timezone-code", type=click.Choice(POSSIBLE_TIMEZONE_CODES), default="America/New_York")
 @click.option("--dbm-query-param-type", default="TYPE_TRUEVIEW")
 @click.option("--dbm-start-date", type=click.DateTime())
 @click.option("--dbm-end-date", type=click.DateTime())
--- a/ack/readers/google_dbm/cli.py
+++ b/ack/readers/google_dbm/cli.py
@@ -35,6 +35,8 @@ from ack.utils.processor import processor
 @click.option("--dbm-query-title")
 @click.option("--dbm-query-frequency", type=click.Choice(POSSIBLE_FREQUENCIES), default="ONE_TIME")
 @click.option("--dbm-query-timezone-code", type=click.Choice(POSSIBLE_TIMEZONE_CODES), default="America/New_York")
+@click.option("--dbm-scheduled-start-date", type=click.DateTime())
+@click.option("--dbm-scheduled-end-date", type=click.DateTime())
 @click.option("--dbm-query-param-type", default="TYPE_TRUEVIEW")
 @click.option("--dbm-start-date", type=click.DateTime())
 @click.option("--dbm-end-date", type=click.DateTime())
--- a/ack/readers/google_dbm/config.py
+++ b/ack/readers/google_dbm/config.py
@@ -61,6 +61,9 @@ class GoogleDBMReaderConfig(BaseModel):
     query_id: str = None
     query_title: str = None
     query_frequency: str = "ONE_TIME"
+    query_timezone_code: str = "America/New_York"
+    scheduled_start_date: datetime
+    scheduled_end_date: datetime
     query_param_type: str = "TYPE_TRUEVIEW"
     start_date: datetime = None
     end_date: datetime = None
--- a/ack/readers/google_dbm/config.py
+++ b/ack/readers/google_dbm/config.py
@@ -23,9 +23,26 @@ from pydantic import BaseModel, validator
 GOOGLE_TOKEN_URI = "https://accounts.google.com/o/oauth2/token"
 
 DAY_RANGES = ("PREVIOUS_DAY", "LAST_30_DAYS", "LAST_90_DAYS", "LAST_7_DAYS", "PREVIOUS_MONTH", "PREVIOUS_WEEK")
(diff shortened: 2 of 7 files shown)


ack/readers/google_dbm/reader.py
```diff
@@ -55,9 +55,12 @@ def __init__(self, access_token, refresh_token, client_secret, client_id, **kwar
...
 
-        check_date_range_definition_conformity(
-            self.kwargs.get("start_date"), self.kwargs.get("end_date"), self.kwargs.get("day_range")
-        )
+        if self.kwargs.get("query_frequency") == "ONE_TIME":
+            check_date_range_definition_conformity(
+                self.kwargs.get("start_date"), self.kwargs.get("end_date"), self.kwargs.get("day_range")
+            )
+        else:
+            check_date_range_definition_conformity(self.kwargs.get("start_date"), self.kwargs.get("end_date"), None)
```
gabrielleberanger: Good job ! 👍 

Quick question : why do we need these two cases here ?

Also, for my understanding : does the start_date and end_date parameters represent the same thing in custom_reports and custom_scheduled_reports ? Or does it represent the period of the queried data in one case, and the period on which the scheduled report should run in the second case ?

ack/readers/google_dbm/cli.py
```diff
@@ -33,7 +33,8 @@
 @click.option("--dbm-request-type", type=click.Choice(POSSIBLE_REQUEST_TYPES), required=True)
 @click.option("--dbm-query-id")
 @click.option("--dbm-query-title")
-@click.option("--dbm-query-frequency", default="ONE_TIME")
+@click.option("--dbm-query-frequency", type=click.Choice(POSSIBLE_FREQUENCIES), default="ONE_TIME")
+@click.option("--dbm-query-timezone-code", type=click.Choice(POSSIBLE_TIMEZONE_CODES), default="America/New_York")
```
gabrielleberanger, line 37: Could you add this new option to the [documentation](https://artefactory.github.io/artefactory-connectors-kit/readers.html#google-doubleclick-bid-manager-reader-dbm) ? A small guide on how to use sphinx is available [here](https://artefactory.github.io/artefactory-connectors-kit/contributing.html#documentation).

ack/readers/google_dbm/reader.py
benoitbazouin: Indeed, start date and end date represent the period of the queried data in one case, and the period on which the scheduled report should run in the second case. It might be better to add two arguments (like scheduled-end-date + scheduled-start-date). 
What do you think ? 

I needed these two cases because when querying a scheduled report, we should be able to have both a day-range (Last_7_Days for example) and a couple (start-date, end-date).

ack/readers/google_dbm/cli.py
benoitbazouin: Yes thank you

ack/readers/google_dbm/reader.py
gabrielleberanger: Yes, it might be clearer to have dedicated date arguments for the custom_scheduled_report 😊 Otherwise it can be quite confusing.

Got it for the 2 cases! Then shouldn't we add a check to verify that the user did not provide a day_range if the request type is custom_scheduled_report (because it does not make sense)?

benoitbazouin: In fact what doesn't make sense is having both a day_range and a couple(start_date, end_date) for "normal" reports.

But for scheduled reports, we need to have both a day_range and (schedule_start_date, schedule_end_date): 
query on "LAST_7_DAYS" (day-range) between 07/07/21 (schedule-start-date) and 07/08/21 (schedule-end-date) on a "DAILY" basis (=frequency) for example. Thus I need to create a specific check for scheduled reports (different from check_date_range_definition_conformity)

ack/readers/google_dbm/config.py
```diff
@@ -44,6 +61,9 @@ class GoogleDBMReaderConfig(BaseModel):
     query_id: str = None
     query_title: str = None
     query_frequency: str = "ONE_TIME"
+    query_timezone_code: str = "America/New_York"
```
pol-defont-reaulx: By writing "Literal[tuple(POSSIBLE_TIMEZONE_CODES)]" instead of "str" you can impose that the value given to the variable has to be in the list you wrote before
Same thing for frequency as you wrote the list POSSIBLE_FREQUENCIES just before

docs/source/readers.rst
```diff
@@ -724,10 +724,13 @@ CMD Options                     JSON Options                 Definition
...
 ``--dbm-refresh-token``         ``refresh_token``            Refresh token for OAuth2
-``--dbm-query-request-type``    ``query_request_type``       Doubleclick Bid Manager API request type. Possible values: existing_query, custom_query, existing_query_report, custom_query_report, lineitems_objects, sdf_objects and list_reports.
+``--dbm-query-request-type``    ``query_request_type``       Doubleclick Bid Manager API request type. Possible values: existing_query, custom_query, existing_query_report, custom_query_report, custom_scheduled_query, lineitems_objects, sdf_objects and list_reports.
 ``--dbm-query-id``              ``query_id``                 Query ID.
 ``--dbm-query-title``           ``query_title``              Query title, used to name the reports generated from this query in DV360 UI.
 ``--dbm-query-frequency``       ``query_frequency``          How often the query is run. Possible values can be found `here <https://developers.google.com/bid-manager/v1/queries#schedule.frequency>`__. Default: ONE_TIME.
+``--dbm-query-timezone-code``   ``query_timezone_code``      Canonical timezone code for report generation time. Defaults to America/New_York.
```
pol-defont-reaulx, line 731: If you update the doc for cli you should also update it for json entrypoint, which is just bellow this one I think

ack/utils/date_handler.py
```diff
@@ -69,6 +69,24 @@ def check_date_range_definition_conformity(start_date: date, end_date: date, dat
             raise DateDefinitionException("Report end date should be equal or ulterior to report start date.")
 
 
+def check_scheduled_parameters_definition_conformity(
+    scheduled_start_date: date, scheduled_end_date: date, frequency: str, date_range: str
+):
+
+    if not date_range:
```
pol-defont-reaulx, line 76: From what I see it's almost the same code as check_date_range_definition_conformity, is it possible to call this function from the new one by doing the new tests before calling check_date_range_definition_conformity? It should avoid code duplication

pol-defont-reaulx (changes requested):

ack/readers/google_dbm/config.py
benoitbazouin: ok thanks

docs/source/readers.rst
benoitbazouin: it is already included here (column JSON options), or am I missing something ?

ack/utils/date_handler.py
benoitbazouin: The logic is not the same for a scheduled report and for a normal report. 
In a normal report, we shouldn't have both a date range and a couple of dates. 
In a scheduled report, it is needed. 

I could indeed use check_date_range_definition_conformity(scheduled_start_date, scheduled_end_date, None) after my else, but it seems to me that it is a "hack" just to avoid code deduplication that doesn't explain the logic behind the check. 

What do you think @AlexisVLRT ?

docs/source/readers.rst
pol-defont-reaulx: ah yes my bad sorry

ack/utils/date_handler.py
pol-defont-reaulx: It seems logical to me, thanks for the explanation

pol-defont-reaulx (approved):
lgtm

PR #160, opened 2021-08-02
Added functionality for filter criteria checks

bdavis9725:
### Issue

- [x] This PR handles a dimension filter check, if the report is not of the 'Standard' type then the dimension filters method will throw a KeyError. This code solves this by checking the report type and adding the appropriate key into the dictionary.

### Description

- [x] Adds the appropriate criteria name for the dictionary dependant on report type for dcm.

### Tests

- [x] This is a conditional statement, any issues will be updated, but functionality should be successful.


bibimorlet (approved):
Please ensure that you're able to query reports with these modifications.
After that you can merge the Pull Request.

PR #164, opened 2021-08-30
fix linting issues and tests google_dcm

benoitbazouin:
Fix linting issues for Google DCM client.
Fix and add tests according to latest update


gabrielleberanger (approved):
Just for my knowledge : the latest update you're talking about is that the Google DCM connector now accepts two types of reports (standard and reach), right ?

PR #165, opened 2021-08-30
fix github action GCP deploy image

benoitbazouin:
Fix CD deploy image


gabrielleberanger (approved):
Merci @benoitbazouin !

PR #168, opened 2021-09-29
update/twitter-ads-requirement

bdavis9725:
Small change to update the Twitter Ads version in the requirements.txt. Version 8.0 is no more, current version is now Version 9.0


bibimorlet (approved):
If the reports still work with this version, LGTM 👍

PR #169, opened 2021-10-05
Update-SearchConsole-API-Reader

bdavis9725:
### Issue

The Search Console API has been updated, and thus a change to facilitate this change is required, this has been implemented in this request as per their documentation.

### Description

This change is made to make sure the codebase complies with the changes outlined in the API documentation.
Please see https://developers.google.com/search/blog/2020/12/search-console-api-updates#discovery-doc-migration for further information.

### DIFF
--- a/ack/readers/google_search_console/reader.py
+++ b/ack/readers/google_search_console/reader.py
@@ -78,7 +78,7 @@ class GoogleSearchConsoleReader(Reader):
             http = credentials.authorize(httplib2.Http())
             credentials.refresh(http)
 
-            self._service = build(serviceName="webmasters", version="v3", credentials=credentials, cache_discovery=False)
+            self._service = build(serviceName="searchconsole", version="v1", credentials=credentials, cache_discovery=False)
 
     @staticmethod
     def check_end_date(end_date):
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,3 @@
--r requirements.txt
 black==19.10b0
 flake8==3.8.3
 freezegun==0.3.15
--- a/.github/workflows/deploy_doc.yaml
+++ b/.github/workflows/deploy_doc.yaml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python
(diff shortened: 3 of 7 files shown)


bibimorlet:
The checks didn't pass. The build couldn't be done because of pip requirements issues.

requirements-dev.txt
```diff
@@ -1,4 +1,3 @@
--r requirements.txt
```
bibimorlet, line 1: Why is this line removed?

mycaule: I removed it in the other PR it because the workflow did it in two steps.

https://github.com/artefactory/artefactory-connectors-kit/blob/c16cda81d45a4c1f9f885ad87b138716adf02c32/.github/workflows/lint_and_run_tests.yml#L42-L43

bibimorlet: Ok so there was useless redundancy am I right?

mycaule: I think the way includes in requirements.txt files are meat to be used is you define parent requirements files and derives other children files and just install the childrens.

In that case it would be just one commands:
```
# For the CI/CD (Google Actions, lint_and_run_tests.yml)
pip install -r requirements-dev.txt --upgrade --no-cache-dir --use-deprecated=legacy-resolver

# For production (Docker image, Dockerfile)
pip install -r requirements.txt --upgrade --no-cache-dir --use-deprecated=legacy-resolver
```

ack/readers/google_search_console/reader.py
```diff
@@ -78,7 +78,7 @@ def initialize_analyticsreporting(self):
             http = credentials.authorize(httplib2.Http())
             credentials.refresh(http)
 
-            self._service = build(serviceName="webmasters", version="v3", credentials=credentials, cache_discovery=False)
+            self._service = build(serviceName="searchconsole", version="v1", credentials=credentials, cache_discovery=False)
```
mycaule, line 81: LGTM @bibimorlet.
Also upgrading the dependency to `google-api-python-client==2.29.0` would be great.

bibimorlet (approved):

PR #170, opened 2021-10-12
Build errors with new version of pip and MacOS

mycaule:
- Fixes the error that occurs when running `pip install googleanalytics` which depends on a old version of `keyring` `ERROR: No matching distribution found for keyring==5.3`
- Bump `numpy` to a newer version to be able to support compilation with ARM chips on Mac OS.
- Cleanup the `Dockerfile` and one GitHub Actions workflow : `gcc` was needed at some point during the global `pip install`

References: https://stackoverflow.com/questions/67074684/pip-has-problems-with-metadata

### DIFF
--- a/.github/workflows/deploy_doc.yaml
+++ b/.github/workflows/deploy_doc.yaml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v1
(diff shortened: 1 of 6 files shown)


mycaule:
This should also fix your build on #169 @BrianALDavis.

I basically reproduced the error both on my machine and on a GitHub Actions VM.

bibimorlet (approved):

PR #178, opened 2022-01-18
Added awin reader code

bdavis9725:
### Description

Added code to enable the AWIN Advertiser API

### Tests

Tests done locally, the outcome is as expected from the API, with no errors


ack/entrypoints/json/readers.py
```diff
@@ -36,6 +37,7 @@
 from ack.readers.adobe_analytics_1_4.reader import AdobeAnalytics14Reader
 from ack.readers.adobe_analytics_2_0.reader import AdobeAnalytics20Reader
 from ack.readers.amazon_s3.reader import AmazonS3Reader
+from ack.readers.awin_advertiser import AwinAdvertiserReader
```
pol-defont-reaulx: you're missing a `.reader` in your `from` statement

pol-defont-reaulx (changes requested):
I don't think it's why it's not working but still a mini error.

ack/readers/awin_advertiser/cli.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@click.command(name="read_awin")
+@click.option("--awin-auth-token", required=True)
+@click.option("--awin-advertiser-id", required=True)
+@click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
+@click.option("--awin-region", required=True)
+@click.option("--awin-timezone", required=True)
+@click.option("--awin-start-date", type=click.DateTime())
+@click.option("--awin-end-date", default=None, type=click.DateTime())
```
pol-defont-reaulx, line 33: You're missing a `@processor()` decorator here which is why it isn't working. This decorator is made to protect the sensible informations of cmd. Even if there's no sensible information you have to put it but empty. It works as a double call where the first call go through the processor function which protect the sensible informations and return the `awin_advertiser` function.
You should add in the processor the token and maybe the advertiser id if it's sensible. (The field `--awin-auth-token` become `awin_auth_token` after processing by click

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/cli.py
```diff
@@ -0,0 +1,35 @@
...
+def awin_advertiser(**kwargs):
+    return AwinAdvertiserReader(**extract_args("awin_", kwargs))
```
pol-defont-reaulx: You're missing a `@processor()` decorator here which is why it isn't working. This decorator is made to protect the sensible informations of cmd. Even if there's no sensible information you have to put it but empty. It works as a double call where the first call go through the processor function which protect the sensible informations and return the `awin_advertiser` function.
You should add in the processor the token and maybe the advertiser id if it's sensible. (The field `--awin-auth-token` become `awin_auth_token` after processing by click

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/cli.py
bdavis9725: This change has been made, will be added to the next commit

ack/readers/awin_advertiser/reader.py
```diff
@@ -0,0 +1,81 @@
...
+        timezone = self.timezone
+        start_date = self.start_date
+        end_date = self.end_date
+
+        if report_type == REPORT_TYPES[0]:
+            build_url = f"{AWIN_API_ENDPOINT}{advertiser_id}/reports/publisher?startDate={start_date}&endDate={end_date}&timezone={timezone}&accessToken={auth_token}"
+        elif report_type == REPORT_TYPES[1]:
+            build_url = f"{AWIN_API_ENDPOINT}{advertiser_id}/reports/creative?startDate={start_date}&endDate={end_date}&region={region}&timezone={timezone}&accessToken={auth_token}"
+        elif report_type == REPORT_TYPES[2]:
+            build_url = f"{AWIN_API_ENDPOINT}{advertiser_id}/reports/campaign?startDate={start_date}&endDate={end_date}&accessToken={auth_token}"
```
louisRDSC: I think a simple 
`build_url = f"{AWIN_API_ENDPOINT}{advertiser_id}/reports/{self.report_type}?startDate={start_date}&endDate={end_date}&timezone={timezone}&accessToken={auth_token}" `

ack/readers/awin_advertiser/reader.py
```diff
@@ -0,0 +1,81 @@
...
+        
+    @retry
+    def request(self):
+        auth_token = self.auth_token
+        advertiser_id = self.advertiser_id
+        report_type = self.report_type
+        region = self.region
+        timezone = self.timezone
+        start_date = self.start_date
+        end_date = self.end_date
```
louisRDSC: everything is already available in self, it is quite redundant to redeclare all the variables

ack/readers/awin_advertiser/reader.py
```diff
@@ -0,0 +1,81 @@
...
+        response = requests.get(
+            build_url
+        )
+        json_response = response.json()
+        logger.debug(f"Response: {json_response}")
+        return json_response
+
+    @staticmethod
+    def create_date_range(start_date, end_date):
+        return {"min": start_date.strftime(DATEFORMAT), "max": end_date.strftime(DATEFORMAT)}
```
louisRDSC: create_date_range is not proper to awin reader and use class properties thus it should not be a private method, I think being a private method would be more appropriate ( or maybe you could use https://github.com/artefactory/artefactory-connectors-kit/blob/ac821f8d58fd9569f75e1df3a9c017f6d3a1f5f3/ack/utils/date_handler.py#L86)

louisRDSC (changes requested):

ack/readers/awin_advertiser/reader.py
bdavis9725: This is tricky because each API call requires different elements within the URL. For e.g. you cannot use region in a publisher report but you can use this in a creative report.

ack/readers/awin_advertiser/reader.py
bdavis9725: This change has been made in the latest code.

ack/readers/awin_advertiser/reader.py
bdavis9725: This might not really be needed at all, I will comment this section out for now.

ack/readers/awin_advertiser/reader.py
louisRDSC: I see , It may be good to at least build the common part outside. also the documentation here can be helpful 
https://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls

.env
```diff
@@ -1,4 +1,4 @@
 PROJECT_ID=artefact-docker-containers
 DOCKER_IMAGE=artefactory-connectors-kit-dev
-DOCKER_TAG=v2.0
+DOCKER_TAG=v2.1-awinreaderaddition
```
pol-defont-reaulx: 2.1 only would be good for the docker tag

ack/readers/awin_advertiser/config.py
```diff
@@ -0,0 +1,55 @@
...
+class AwinAdvertiserReaderConfig(BaseModel):
+    auth_token: str
+    advertiser_id: str
+    report_type: str
+    region: str
+    timezone: str
+    start_date: datetime = None
+    end_date: datetime = None
+    date_range: Tuple[datetime, datetime] = None
+    # day_range: Literal[DAY_RANGE] = None
```
pol-defont-reaulx: If you want to implement a data range, it's already done in data handler I think to have the last 7 days, last month or else.

ack/readers/awin_advertiser/reader.py
```diff
@@ -40,20 +40,32 @@ def __init__(
...
-            build_url = f"{AWIN_API_ENDPOINT}{self.advertiser_id}"\
-                f"/reports/{self.report_type}?startDate={self.start_date}&endDate={self.end_date}"\
-                f"&timezone={self.timezone}&accessToken={self.auth_token}"
+            payload = {
+                'startDate': self.start_date, 
+                'endDate': self.end_date, 
+                'timezone': self.timezone, 
+                'accessToken': self.auth_token
+                }
+            build_url = f"{AWIN_API_ENDPOINT}{self.advertiser_id}/reports/{self.report_type}?"
```
pol-defont-reaulx: As build_url is common to the three cases so you can make only one declaration outside the if statement, your code will be cleaner ;)

ack/readers/awin_advertiser/reader.py
bdavis9725: This has been changed now to use params, as per the link above

ack/readers/awin_advertiser/config.py
bdavis9725: This is not needed, so it has been removed

docs/source/readers.rst
```diff
@@ -197,6 +197,52 @@ CMD Options                     JSON Options                         Definition
...
+
+---------------
+Command options
+---------------
+
+=========================  =======================  =================================================================================================================================================================================
+CMD Options                JSON Options             Definition
+=========================  =======================  =================================================================================================================================================================================
+``--awin-auth-token``      ``auth-token``           Authorisation Token, which can request through the AWIN API documentation website
+``--awin-advertiser-id``   ``advertiser-id``        Advertiser ID, which you can find on the AWIN Platform
```
pol-defont-reaulx: For the json options here, it's advertiser_id I think

docs/source/readers.rst
```diff
@@ -197,6 +197,52 @@ CMD Options                     JSON Options                         Definition
...
+``--awin-report-type``     ``report-type``          Report Type, which are defined in the Advertiser API documentation
```
pol-defont-reaulx: Same here, it's report_type

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/cli.py
```diff
@@ -0,0 +1,36 @@
...
+
+@click.command(name="read_awin")
+@click.option("--awin-auth-token", required=True)
+@click.option("--awin-advertiser-id", required=True)
+@click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
+@click.option("--awin-region", required=True)
+@click.option("--awin-timezone", required=True)
+@click.option("--awin-start-date", type=click.DateTime())
+@click.option("--awin-end-date", default=None, type=click.DateTime())
+@processor("awin-auth-token", "awin-advertiser-id")
```
pol-defont-reaulx: When passing through the cli library, `-`become `_` for the processor arguments. So `awin-auth-token` should be `awin_auth_token` and same for the others

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/config.py
```diff
@@ -0,0 +1,52 @@
...
+    "publisher",
+    "creative",
+    "campaign"
+]
+
+
+class AwinAdvertiserReaderConfig(BaseModel):
+    auth_token: str
+    advertiser_id: str
+    report_type: str
```
pol-defont-reaulx: For field like this, you can enforce the value to be one of the REPORT_TYPES value. You just have to write `report_type: Literal[tuple(REPORT_TYPES)]`.
I also saw that in the cli file you put a default value for report_type so you can add it here also (if it's logical to you)

ack/readers/awin_advertiser/cli.py
```diff
@@ -0,0 +1,36 @@
...
+from ack.utils.processor import processor
+
+
+@click.command(name="read_awin")
+@click.option("--awin-auth-token", required=True)
+@click.option("--awin-advertiser-id", required=True)
+@click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
+@click.option("--awin-region", required=True)
+@click.option("--awin-timezone", required=True)
+@click.option("--awin-start-date", type=click.DateTime())
```
pol-defont-reaulx: You should add a `default=None` as you did for the end date no?

pol-defont-reaulx (changes requested):

pol-defont-reaulx (approved):
LGTM, well done for this reader and thanks for the contribution ;)
Just wait for @louisRDSC review and you'll be good to merge

louisRDSC (approved):
LGTM , Good Job

ack/readers/awin_advertiser/reader.py
```diff
@@ -0,0 +1,77 @@
...
+                'endDate': self.end_date,
+                'accessToken': self.auth_token
+            }
+
+        response = requests.get(
+            build_url, params=payload
+        )
+        json_response = response.json()
+        logger.debug(f"Response: {json_response}")
+        return json_response
```
mycaule, line 69: I wonder if when authentication fails, the python script can return with an error (Bash exit 1 depending on the HTTP status).
... Just to be catched up by log aggregator when deployed on the cloud.

The current way it is is that it will print a silent error.

PR #179, opened 2022-01-20
Feat/awin updates

bdavis9725:
### Issue

- This update adds functionality to the AWIN reader for campaign reports. It also removes an unnecessary nested JSON key 'tags', which are not required.

### Tests

- Tests done locally to confirm outcomes are as expected, and functionality is unchanged, based on the adjustments.

### Documentation

- I have added the new parameters of awin-campaign and awin-interval to the documentation, however I have also specified that these are not required, as they are only used for campaign reports.


ack/readers/awin_advertiser/cli.py
```diff
@@ -28,7 +28,9 @@
 @click.option("--awin-advertiser-id", required=True)
 @click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
 @click.option("--awin-region", required=True)
+@click.option("--awin-campaign")
 @click.option("--awin-timezone", required=True)
+@click.option("--awin-interval")
```
pol-defont-reaulx: If they are not required, you should put a default value

ack/readers/awin_advertiser/config.py
```diff
@@ -39,7 +39,9 @@ class AwinAdvertiserReaderConfig(BaseModel):
     advertiser_id: str
     report_type: Literal[tuple(REPORT_TYPES)]
     region: str
+    campaign: str
     timezone: str
+    interval: str
```
pol-defont-reaulx: Same here

ack/readers/awin_advertiser/reader.py
```diff
@@ -65,6 +69,10 @@ def request(self):
             build_url, params=payload
         )
         json_response = response.json()
+        # if report is not a campaign report, remove tags from the response
+        if self.report_type != REPORT_TYPES[2]:
+            for element in json_response:
+                element.pop('tags')
```
pol-defont-reaulx: Why do you need to remove tags? Just to understand here

pol-defont-reaulx (changes requested):
Good job, just some modifications to do.
I also saw that I forgot to tell you to update the doc in doc/source/overview.rst. You can add your reader to the list as you did in the readme.

ack/readers/awin_advertiser/cli.py
bdavis9725: have added default=None to each

ack/readers/awin_advertiser/reader.py
bdavis9725: They are not necessary for the use case I am using the AWIN reader for, however if you prefer I keep them on the reader and could create a copy of ACK with this modification instead done outside of the repo. Please let me know.

pol-defont-reaulx: Yes it is better to answer specific needs outside of the official repo. If other people want to use your reader they'll not have the same needs as you so keep it general.
Another solution is to add an boolean argument to your reader like `--remove-tag` for example and to only do your modification if this argument is set to True.

bdavis9725: have made this an optional argument now

ack/readers/awin_advertiser/config.py
```diff
@@ -39,7 +39,10 @@ class AwinAdvertiserReaderConfig(BaseModel):
     advertiser_id: str
     report_type: Literal[tuple(REPORT_TYPES)]
     region: str
+    campaign: str = None
     timezone: str
+    interval: str = None
+    remove_tags: str = 'False'
```
pol-defont-reaulx: remove_tags should be a boolean with a default value at False (without quotes as it's a real boolean)

ack/readers/awin_advertiser/reader.py
```diff
@@ -65,6 +70,10 @@ def request(self):
             build_url, params=payload
         )
         json_response = response.json()
+        # if report is not a campaign report, remove tags from the response
+        if self.remove_tags == 'True' and self.report_type != REPORT_TYPES[2]:
```
pol-defont-reaulx: If you make remove_tags a boolean you just have to put:
`if self.remove_tags and self.report_type != REPORT_TYPES[2]:`

ack/readers/awin_advertiser/cli.py
```diff
@@ -28,7 +28,10 @@
 @click.option("--awin-advertiser-id", required=True)
 @click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
 @click.option("--awin-region", required=True)
+@click.option("--awin-campaign", default=None)
 @click.option("--awin-timezone", required=True)
+@click.option("--awin-interval", default=None)
+@click.option("--awin-remove-tags", default='False')
```
pol-defont-reaulx: Same here, make it a real boolean by removing the quotes: `default=False`

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/reader.py
bdavis9725: done

ack/readers/awin_advertiser/cli.py
bdavis9725: done

ack/readers/awin_advertiser/config.py
bdavis9725: done

ack/readers/awin_advertiser/cli.py
```diff
@@ -28,7 +28,10 @@
 @click.option("--awin-advertiser-id", required=True)
 @click.option("--awin-report-type", type=click.Choice(REPORT_TYPES), default=REPORT_TYPES[0])
 @click.option("--awin-region", required=True)
+@click.option("--awin-campaign", default=None)
 @click.option("--awin-timezone", required=True)
+@click.option("--awin-interval", default=None)
+@click.option("--awin-remove-tags", default=False)
```
pol-defont-reaulx: I think you should specify that the type here is boolean. It can avoid errors if people make mistakes. It's just a detail though

ack/readers/awin_advertiser/reader.py
pol-defont-reaulx: thanks

ack/readers/awin_advertiser/cli.py
bdavis9725: this makes sense, yes, I have done this quickly

ack/readers/awin_advertiser/cli.py
```diff
@@ -31,7 +32,7 @@
 @click.option("--awin-campaign", default=None)
 @click.option("--awin-timezone", required=True)
 @click.option("--awin-interval", default=None)
-@click.option("--awin-remove-tags", default=False)
+@click.option("--awin-remove-tags", default=False, type=Boolean)
```
pol-defont-reaulx: You should use the type that is build with Click: `type=click.BOOL` and not import a boolean from another library ;)

pol-defont-reaulx (changes requested):

ack/readers/awin_advertiser/cli.py
bdavis9725: ahaha I didn't even see that it did that

pol-defont-reaulx (approved):
LGTM, good job ;)

PR #180, opened 2022-01-21
Feat/awin add date to api

bdavis9725:
### Issue

- The API doesn't output a date, this can be troublesome for Berlioz, which works closely with ACK.

### Description

- This PR adds the start_date from the reader CLI request, only if the start_date and end_date match, and only if the date is not part of the json response from the API.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
 PROJECT_ID=artefact-docker-containers
 DOCKER_IMAGE=artefactory-connectors-kit-dev
-DOCKER_TAG=v2.3
+DOCKER_TAG=v2.1.2
 DOCKER_REGISTRY=eu.gcr.io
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -1,4 +1,4 @@
 PROJECT_ID=artefact-docker-containers
 DOCKER_IMAGE=artefactory-connectors-kit-dev
-DOCKER_TAG=v2.1
+DOCKER_TAG=v2.3
```
pol-defont-reaulx: It's just some minor corrections, you can let the docker_tag to 2.1. Or maybe 2.1.2 if you really want to increase the version ;)

bdavis9725: Sure! I have made this change now.

pol-defont-reaulx (approved):
LGTM

